Issue #19, opened 2014-05-14
Suggestion: Combine directories of they have only one subdirectory


W4RH4WK:
hi there, great work so far
my suggestion would be combining multiple directories if they only have one subdirectory, for example:

```
- src
  `- main
    `- java
      `- at
        `- bluephoenix
          `- Package
            `- Main.java
            `- Part.java
```

would become

```
- src/main/javas/at/bluephoenix/Package
  `- Main.java
  `- Part.java
```

PR #48, opened 2014-05-14
Collapse directories with only single subdirectory

chalup:
Implements #19

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -101,7 +101,21 @@
         // TOOD: handle submodule
       })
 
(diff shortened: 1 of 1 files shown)


Amit-Kumar-G:
Hey @buunguyen , I see that feature is merged, but cannot find a setting to enable this? Could someone point me to where it is? Thanks :)

buunguyen:
It's the default behavior, no setting. Note that if you enable lazy loading for repositories, this won't work though.

PR #116, opened 2014-06-16
GitLab support

lynxaegon:
I left the octicons from the cdn, because i know how to load a local file just for chrome. 
GitLab is working, adapters change based on an adapterManager that returns the correct adapter.
There are some minor hiccups while moving through a gitlab repo, sometimes the whole page reloads, and atm it loading from cache / live data takes a few seconds, i think it has something to do with jstree, i'll check these days.


albacoretuna:
This is a new extension specifically for Gitlab if you came here by searching: 
https://github.com/tavyandy97/span-tree

Issue #151, opened 2015-01-13
Octotree breaks appearance of pull request diffs


ruzz311:
I keep getting an issue when I navigate to Pull Requests after having octotree open, specifically when I'm using split diffs.  The issue dissappears when I refresh the diff screen.  

[Video capture of bug happening](https://cl.ly/ZGdE?_ga=1.129915531.609429140.1421164867)

![PR display issue Screenshot](http://f.cl.ly/items/210r422H2B3w061I3n0B/Screen%20Shot%202015-01-13%20at%2010.06.50%20AM.png)

I've also had the display of the repo's main page have weird negative margin where the content is no longer centered, but I'm having a hard time recreating it.  It must be a very specific sequence of events.

buunguyen:
@Ephemera can you help looking into this?

Ephemera:
@buunguyen Ok. I will try to fix it this weekend.

buunguyen:
@Ephemera Thank you, much appreciated!

PR #156, opened 2015-01-21
Remove margin when sidebar is hidden. Fixes #155.

### LINKED ISSUE #155: Left Margin added to container breaks ZenHub layout
I reported this [issue at the ZenHub repository](https://github.com/ZenHubIO/support/issues/185), but discovered it was this Chrome extension causing the problems.

ZenHub fills the full browser window width and is centered using `auto` margins. When I resize my browser window Octotree adds a margin right to the container and shifts part of ZenHub offscreen.

[screenr](https://www.screenr.com/WluN)

2015-01-20 18_43_12-boards zenhubio_support

Mottie:


Mottie:
Oh, did you test this? It works perfectly for me & I think this solution is much cleaner than the other pull request... but you're the boss here.

buunguyen:
Actually the other commit addresses another issue (see #151), incidentally taking care of this one as well.

buunguyen:
Oh, I see what you meant. Now it's getting interesting. Your solution is indeed very simple. The other is more complex but handle the split-view more nicely (i.e. not having to scroll to view diff). Tough decision...

PR #174, opened 2015-07-03
Try to detect tags as well as branches and commits.

drusso:
Octotree will fallback to rendering the `master` branch's tree when a user is actually viewing/navigating a tag. I am using Octotree v1.6.3 against the latest github.com. 

This happens because the branch/tag dropdown does not contain the icon class `octicon-git-branch` when a tag is selected. Instead it contains the `octicon-tag` icon class. This PR expands the selector to look for either `.octicon-git-branch` or `.octicon-tag`.

With this fix, Octotree will successfully detect and render a tag's tree.

### DIFF
--- a/src/adapter.github.js
+++ b/src/adapter.github.js
@@ -7,7 +7,7 @@ const
       'search', 'developer', 'account'
     ]
   , GH_RESERVED_REPO_NAMES = ['followers', 'following', 'repositories']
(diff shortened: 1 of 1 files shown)


Ephemera:
I think `octicon-git-branch` and `octicon-tag` icon class does not work any more for that. It's just selecting the icon in the overall-summary bar.

PR #184, opened 2015-09-05
Load individual folder

crashbell:
I had to modify view.tree.js by adding a dummy function to work around with this lazy load. Still don't find any good way to make this bad code better. 

```javascript
TreeView.prototype.fetchData = function(opts, success) {
  // Dummy func
  throw "It must be implemented from caller when initializing this object"
}
```

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -175,7 +175,7 @@ TreeView.prototype.syncSelection = function(currentPath) {
     }
     else
       // 300ms seems enough to fetch new items
-      // it will repeat if there is no node found
+      // repeats if there is no new node fetched
       setTimeout(function(){
         self.syncSelection(path)
       }, 300)
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -74,23 +74,25 @@ TreeView.prototype.showHeader = function(repo) {
     })
 }
 
-TreeView.prototype.show = function(repo, token, treeData) {
+TreeView.prototype.show = function(repo, token) {
   var self = this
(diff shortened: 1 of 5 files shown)


src/view.tree.js
```diff
@@ -99,19 +113,53 @@ TreeView.prototype.show = function(repo, treeData) {
...
-  tree.deselect_all()
-
-  // e.g. converts /buunguyen/octotree/type/branch/path to path
-  var match = path.match(/(?:[^\/]+\/){4}(.*)/)
-    , nodeId
-  if (match) {
-    nodeId = PREFIX + decodeURIComponent(match[1])
-    tree.select_node(nodeId)
-    tree.open_node(nodeId)
+
```
crashbell: The code below is for loading nodes continuously until getting to selected node in GitHub URL.

buunguyen:
@crashbell this is awesome! Thank you. I'll be reviewing behaviors/code and commenting here. 

First issue: enable the lazy load option, once the tree is loaded, enable the collapse folder option and save => the spinner will spin forever as there's an error. Note that updates to collapse-folder option or token will trigger a `tryLoadRepo(force=true)`.

buunguyen:
Second: can you rebase and reissue a PR with a single commit? Much easier to review that way.

buunguyen:
In `octotree#optionsChanged()`, you should set `reload` to true whenever the lazy-load option is changed, so that it has immediate effect.

crashbell:
Thanks for reviewing this. I will make a single commit and revise my code regarding your feedback.

PR #187, opened 2015-09-20
add lazy-load option

crashbell:
Regarding your feedback, I made some changes here. I also disabled collapse folder option as discussed. Please take a look and let me know if you see any problem with my code. Thanks!

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -74,15 +74,30 @@ TreeView.prototype.showHeader = function(repo) {
     })
 }
 
-TreeView.prototype.show = function(repo, treeData) {
+TreeView.prototype.show = function(repo, token) {
   var self = this
     , treeContainer = self.$view.find('.octotree_view_body')
     , tree = treeContainer.jstree(true)
     , collapseTree = self.store.get(STORE.COLLAPSE)
+    , recursiveLoad = self.store.get(STORE.RECURSIVE)
 
-  treeData = sort(treeData)
-  if (collapseTree) treeData = collapse(treeData)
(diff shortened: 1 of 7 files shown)


src/view.tree.js
```diff
@@ -119,17 +134,43 @@ TreeView.prototype.show = function(repo, treeData) {
 
 TreeView.prototype.syncSelection = function() {
   var tree = this.$view.find('.octotree_view_body').jstree(true)
-    , path = location.pathname
+    , path = decodeURIComponent(location.pathname)
+    , recursiveLoad = this.store.get(STORE.RECURSIVE)
```
crashbell: duplicated `recursiveLoad = this.store.get(STORE.RECURSIVE)` with function `show` but both of them will be called from `octotree.js` and the store value `RECURSIVE` can be changed after saving options in setting so I decided to make this duplication.

buunguyen: Sure, this is fine.

src/view.tree.js
```diff
@@ -119,17 +134,43 @@ TreeView.prototype.show = function(repo, treeData) {
...
+      var path = (lastPath + "/" + curPath) 
+      arrResult.push(path)
+      return path
+    })
+    return arrResult
   }
+
+  var paths = recursiveLoad ? [currentPath] : createPaths(currentPath)
+
+  function openPathAtIndex (paths, index) {
```
buunguyen: :+1: this version looks better.

crashbell:
I closed this PR to create a new one regarding your feedback

PR #190, opened 2015-10-03
Add Lazy Load Option

crashbell:
Hi Anh,

I updated the code as your comments. I also rebased to your latest code, so it causes new commits when I create this PR. To make all versions consistent, I bumped up all versions in `package.json` and `Info.plist` files.


buunguyen:
:+1:

crashbell:
thanks anh :)

PR #195, opened 2015-10-19
Fix: page is still blank after browser launched

duylam:
Root cause: the driver object from factory is returned (via callback) while the browser app hasn't been ready yet

Solution: wait for 5 seconds


buunguyen:
@duylam :tada:

PR #197, opened 2015-10-26
add more reserved usernames

Justineo:
Hi, octotree is a great extension. When I started to work on my own GitHub extension ([GitHub Hovercard](https://github.com/Justineo/github-hovercard)), I learnt quite a lot from your code.

After digging around GitHub website, I found the reserved usernames should include more and sometimes we may ran into some non-reserved tokens that are not usernames (.eg `password_resest`). I used a regex to filter out invalid usernames in my own extension (`/^[a-z0-9]+$|^[a-z0-9](?:[a-z0-9](?!--)|-(?!-))*[a-z0-9]$/i`). Hope this helps.

### DIFF
--- a/src/adapter.github.js
+++ b/src/adapter.github.js
@@ -4,7 +4,10 @@ const
       'site', 'blog', 'about', 'explore',
       'styleguide', 'showcases', 'trending',
       'stars', 'dashboard', 'notifications',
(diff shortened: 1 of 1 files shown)


buunguyen:
Thanks!

PR #198, opened 2015-10-28
fix branch text too long problem

FrankFan:
when broswe the long-name branch, the branch text will wrap to the second line.this request fix this problem.

before

to

after


buunguyen:
If have a comment. Also, please reissue a clean commit without the merge.

FrankFan:
OK, thank you for your reply. I'll send you a new pull request later.

PR #200, opened 2015-11-08
Support GitLab

crashbell:
I changed a lot and it's really hard to merge it to current branch v2. I created 2 commits, first is for the implementation and second is for merging. 

I moved adapters to `src/adapters` and create a base class for them to share as much code as I can. 
The CSS was splitted to 2 parts for 2 websites and moved to `src/css`

I also moved config files to `src/config`


crashbell:
I close this PR because just found out that it's not working properly with GL enterpise.

PR #202, opened 2015-11-10
Support GitLab

crashbell:
I add new adapter to support GitLab. There are some major changes on CSS and JS structure. README.md was also updated for GitLab access token.


dosire:
@crashbell Thanks so much for supporting GitLab!

PR #203, opened 2015-11-10
added option to enable/disable downloading of file

jamez14:
I found myself accidentally clicking the "download" icon when trying to navigate to a file. Therefore I created an option to remove this option.

* Added an option to the *Settings* panel labeled "Enable downloading of files"
* Enabling (checking) will allow user to download individual files as normal by clicking the blob icon
* Disabling (unchecking) will prevent user from being able to download individual files by clicking the blob icon

Please let me know if you'd like me to change anything (like the location of different functions I created).

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -33,6 +33,7 @@ const
     POPUP    : false,
     SHOWN    : false,
     NONCODE  : false,
+    DOWNLOAD : true
   }
 
   , EVENT = {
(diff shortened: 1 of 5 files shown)


buunguyen:
Sorry, for now I'm inclined not to have another option as it makes Octotree more complicated. I think the current download behavior isn't terribly confusing as there's a download icon upon hover, mistakes may happen occasionally for new users but I think they'll learn to avoid it :).

PR #214, opened 2015-11-25
Add new selector for Github new design

crashbell:
I add new selector to find the current branch and fix error when `match` an empty string.


crashbell:
@buunguyen I updated the commit: remove _detectLocationChange, use MutationObserver instead.

PR #225, opened 2015-12-22
Fix testcases for v2

Ephemera:


test/pageobject.js
```diff
@@ -98,7 +98,7 @@ var controls = {
   ghBreadcrumb    : '.breadcrumb .final-path',
   ghSearch        : '.js-site-search-field',
 
-  helpPopup       : '.octotree_popup',
+  helpPopup       : '.popup',
```
Ephemera: Which one is correct?

PR #228, opened 2015-12-25
GitLab 7.5.3 b656b85: fix display unfriendly

Brightcells:
* After install from [Chrome Web Store](https://chrome.google.com/webstore/detail/octotree/bkhaagjahfmjljalopjnoealnfndnagc/related), it display unfriendly on GitLab 7.5.3 b656b85
image

* Fixed it by adjusting margin-left of '.container>.content' and '.main-nav' 
image

### DIFF
--- a/src/adapters/gitlab.js
+++ b/src/adapters/gitlab.js
@@ -6,6 +6,8 @@ const GL_RESERVED_REPO_NAMES = []
 const GL_HEADER = '.navbar-gitlab'
 const GL_SIDEBAR = '.sidebar-wrapper'
 const GL_TITLE = 'h1.title'
+const GL_CONTENT = '.container>.content'
+const GL_MAIN_NAV = '.main-nav'
 
 class GitLab extends Adapter {
 
@@ -89,6 +91,8 @@ class GitLab extends Adapter {
       $(GL_HEADER).css('z-index', 3)
       $(GL_SIDEBAR).css('z-index', 1)
       $(GL_TITLE).css('margin-left',  sidebarVisible ? '' : 56)
+      // GL_CONTENT and GL_MAIN_NAV is Special for earlier GitLab such as version 7.5.3 b656b85
+      $(GL_CONTENT).parent().add(GL_MAIN_NAV).css('margin-left', sidebarVisible ? 230 : '')
       $('.octotree_toggle').css({
         right: sidebarVisible ? '' : -102,
         top: sidebarVisible ? '' : 8
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 tmp
 .idea
+.DS_Store


buunguyen:
@crashbell can you help verify this pr?

crashbell:
I can't access my GitLab enterprise now hence I have only verified that it won't change anything with latest GitLab version. GitLab stopped supporting 7.5.3 (the oldest is 7.11.2 - link [here](https://packages.gitlab.com/gitlab/gitlab-ee)) so I will verify it with any 7.x I can have.

crashbell:
I have checked with 7.11.2, this PR seems not affect to this version (GL_CONTENT and GL_MAIN_NAV selectors were no longer available). I don't know if we should apply this when it doesn't make any change, even with earliest version GitLab supports so far. 

Note: I found out the logo was missing in 7.x version, I will try to fix that issue.

Brightcells:
GitLab 7.5.3 Released in December 5, 2014 (link [here](https://about.gitlab.com/2014/12/05/gitlab-7-5-3-released/)), just a year ago. Although GitLab stopped supporting earlier versions, I think some people still do not upgrade.

buunguyen:
Can you confirm that this fix won't affect 7.11.2? I'm particularly worried about [this line](https://github.com/buunguyen/octotree/pull/228/files#diff-183cbc91ad44b37654f54bf6014707ffR94), what is the consequence running it in 7.11.2?

crashbell:
I can confirm it won't affect. I installed 7.11.2 in my VPS, you can check it here: http://93.188.161.84/root/my-awesome-project.

src/adapters/gitlab.js
```diff
@@ -89,6 +91,8 @@ class GitLab extends Adapter {
       $(GL_HEADER).css('z-index', 3)
       $(GL_SIDEBAR).css('z-index', 1)
       $(GL_TITLE).css('margin-left',  sidebarVisible ? '' : 56)
+      $(GL_CONTENT).parent().css('margin-left', sidebarVisible ? 230 : '')
```
buunguyen: Please refactor 2 lines to $(GL_CONTENT).parent().add(GL_MAIN_NAV).css(...). Also, add a comment to point out the specific GL version that this code targets

Brightcells:
How about fix display unfriendly by adding styles like the new commit. 
image

src/adapters/gitlab.js
Brightcells: I have changed to fix display unfriendly by adding styles as the new commit.
If finally we decide to use this method, I will refactor the code as below.
```
// Special for GitLab 7.5.3 b656b85
$(GL_CONTENT).parent().add(GL_MAIN_NAV).css('margin-left', sidebarVisible ? 230 : '')
```

buunguyen:
Too complicated. I want a just-work experience, instead of expecting users to tweak around with some CSS styles. The original PR is fine as long as it doesn't break the current version and the 7.11 line, which seems to be the case per @crashbell's comment. So just need to see the code refactored and clarified a bit with comment.

buunguyen:
Closed by mistake.

Brightcells:
Done. Remove code of adding CSS styles in Settings. Just refactored and clarified a bit with comment. Please check and merge.

PR #229, opened 2015-12-28
Should be compared with host and protocol both because `GHEURLS` stores protocol too

### LINKED ISSUE #227: GitHub Enterprise 2.4 TypeError $(...).fire not a function
With Octotree enabled on our GitHub Enterprise 2.4 website, this error keeps showing up in Chrome.
It does not happen on first load, but after a link click inside the website.
Navigating like this`https://github.my.company -> https://github.my.company/jarlefosen` recreates the problem every time on my machine.

```
Uncaught NotSupportedError: Failed to execute 'registerElement' on 'Document':
Registration failed for type 'include-fragment'.
A type with that name is already registered.
(anonymous function) @ frameworks-f8473dece7242da6a20d52313634881b3975c52cebaa1e6c38157c0f26185691.js:7
(anonymous function) @ frameworks-f8473dece7242da6a20d52313634881b3975c52cebaa1e6c38157c0f26185691.js:7

Uncaught TypeError: $(...).fire is not a function
```

This makes links unresponsive to clicks, except that it rethrows the same exception again.

```
Uncaught TypeError: $(...).fire is not a function  @ [reference to minified js]
```

**Setup**
Chrome 47.0.2526.106 (64-bit)
Octotree 2.0.1

*Disabling Octotree and the error does not appear.*

**Difference in page functionality?**
Could it be that Github Enterprise uses some kind of webapp with partials instead of loading the entire page on every location change?

Ephemera:
It may fix #227

### DIFF
--- a/src/octotree.js
+++ b/src/octotree.js
@@ -11,7 +11,9 @@ $(document).ready(() => {
   }
 
   function createAdapter() {
-    const githubUrls = store.get(STORE.GHEURLS).split(/\n/).concat('https://github.com')
+    const githubUrls = store.get(STORE.GHEURLS).split(/\n/)
+      .map((url) => url.replace(/(.*?:\/\/[^/]+)(.*)/, '$1'))
+      .concat('https://github.com')
     return ~githubUrls.indexOf(location.protocol + '//' + location.host)
       ? new GitHub(store)
       : new GitLab(store)
--- a/src/octotree.js
+++ b/src/octotree.js
@@ -11,8 +11,8 @@ $(document).ready(() => {
   }
 
   function createAdapter() {
-    const githubUrls = store.get(STORE.GHEURLS).split(/\n/).concat('github.com')
-    return ~githubUrls.indexOf(location.host)
+    const githubUrls = store.get(STORE.GHEURLS).split(/\n/).concat('https://github.com')
+    return ~githubUrls.indexOf(location.protocol + '//' + location.host)
       ? new GitHub(store)
       : new GitLab(store)
   }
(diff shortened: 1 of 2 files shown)


src/octotree.js
```diff
@@ -11,8 +11,10 @@ $(document).ready(() => {
   }
 
   function createAdapter() {
-    const githubUrls = store.get(STORE.GHEURLS).split(/\n/).concat('github.com')
-    return ~githubUrls.indexOf(location.host)
+    const githubUrls = store.get(STORE.GHEURLS).split(/\n/)
+      .map((url) => url.replace(/(.*?:\/\/[^/]+)(.*)/, '$1'))
```
buunguyen: I'm not sure if this is necessary. Before storing, the url is already [normalized](https://github.com/buunguyen/octotree/blob/master/src%2Fconfig%2Fchrome%2Fbackground.js#L44).

src/octotree.js
```diff
@@ -11,8 +11,10 @@ $(document).ready(() => {
...
+      .concat('https://github.com')
+    return ~githubUrls.indexOf(location.protocol + '//' + location.host)
```
buunguyen: I don't understand why the original `githubUrls.indexOf(location.host)` (e.g. `https://whatever.com'.indexOf('whatever.com')`) didn't work?

src/octotree.js
Ephemera: It does not related with normalize.
When user store with tailing slash, it does not work because `location.host` does not contains tailing slash.

src/octotree.js
Ephemera: Because `githubUrls` is array object so `indexOf` does not work (e.g. `['https://whatever.com'].indexOf('whatever.com')`]

buunguyen: You're right, I didn't understand my own code :).

PR #233, opened 2016-01-07
Fix GitLab URL for project/repository and file path

meip:
Two improvements for a robust GitLab adapter.

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -74,7 +74,7 @@ class Adapter {
             }
             else if (type === 'blob') {
               item.a_attr = {
-                href: `/${repo.username}/${repo.reponame}/${type}/${repo.branch}/${encodeURIComponent(path)}`
+                href: `/${repo.username}/${repo.reponame}/${type}/${repo.branch}/${this.getPath(path)}`
               }
             }
             else if (type === 'commit') {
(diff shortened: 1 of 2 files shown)


crashbell:
@meip I tried to use `projectId` but if user hasn't logged in to GitLab, that ID will not be available because search box disappears.

meip:
Thats true. Because of that I did this change with a fallback for non logged in users.

buunguyen:
Can you clarify why we need to do this? I see your comment about Nginx/Apache, but from my testing the tree currently shows correctly. Is there a particular repo or specific usage when it doesn't work?

meip:
Our internal GitLab is quite a bit outdated (GitLab 6.8.1) and with that specific version or the Apache configuration the request with NAMESPACE/PROJECT in URL received a 404.
I can't provide more detailed informations about the real issue because I'm not the admin of our GitLab. But I tested this fix also on gitlab.com repos and it works. Even as a fallback for non logged-in users which in that case uses the NAMESPACE/PROJECT request.

buunguyen:
Merged by mistake and reverted.

Can you clarify what you have to introduce the `getPath`?

meip:
This is a hook method for Adapter subclasses (b. e. concrete Gitlab adapter) to provide more specific path-resolution functionality.

buunguyen:
I understand that part, just not sure why. Is it also related to your specific deployment? Because the current encoding works fine in my testing.

meip:
Yes its related to our "special" deployment. Do we need more improvement on the PR for more robustness? Any ideas?

buunguyen:
Can you add a bit code comment to clarify that and submit another PR?

meip:
Sorry I can't provide more information on this specific behavior. But I think this issue is very related to #244

Issue #243, opened 2016-01-25
GitLab Community Edition support


GodLesZ:
I've tried to use the Extension on a private GitLab (Community) installation and found just one problem.
The `_getTree`, resolving to `_get('/tree...')`, requests just a wrong path.

*Requested*: `/api/v3/projects/<group>/<project-name>/repository/tree`

After a lil research I found the correct one (maybe only for Community Edition?):
`/api/v3/projects/<project-id>/repository/tree`

The `project-id` kann be found in the page source using `$('#project_id').val()`.
Tested using simple XHR breakpoints and manipulating the request url - works just fine.

I'll create a PR for this but I don't have access to a GitLab Enterprise Edition to test the new behavior on that.
On the gitlab.com page itself for example this is also bugged and my hacky test also works.

Any thoughts?

lusitania:
+1

buunguyen:
Hmm, that's strange. The GitLab folks advised us to use the group/project-name convention because project-id isn't available in the DOM unless you login. I guess a solution is use project ID when it's available but fall back to group/project-name. That wouldn't help in the case when user doesn't login a community installation. I think it's worth reporting the issue to GitLab so that they can address.

A PR is appreciated.

GodLesZ:
@buunguyen you a right - I've just ignored my login state on gitlab.com.
Without logging in the `project_id` is mising on both private and public GitLab instance.

I'll drop an question to the GitLab folks then and try playing around with that fallback idea.
Thank you

buunguyen:
Fixed in 2.0.8.

Spaider:
I have no groups whatsoever and thus looks like Octotree is trying to fetch incorrect URL, which returns HTTP 404. Bad URLs looks something like this `https://gitlab.com/api/v3/projects/<fname>.<lname>%2fitsp-site/repository/tree` (I had login with my first and second names separated by dot)

PR #244, opened 2016-01-25
Optional fallback on project_id for repo fetching if present

GodLesZ:
Like mentioned in #243 this implements fetching from `GitLab` using the `project_id` instead of `group/project-name`.
This seems to working only if the user is logged in.

Tested on both, private GitLab (CE) installation and public gitlab.com (EP) website.

### DIFF
--- a/src/adapters/gitlab.js
+++ b/src/adapters/gitlab.js
@@ -147,7 +147,13 @@ class GitLab extends Adapter {
       // Default from cache
       this._defaultBranch[username + '/' + reponame]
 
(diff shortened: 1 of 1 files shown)


buunguyen:
I'd prefer a simpler change in one place:

```js
const projectId = ....
const project = projectId || (repo.username + '%2f' + repo.reponame)}
const url = `${host}/projects/${project}/repository${path}&private_token=${opts.token}`
```

meip:
Is this PR going to be accepted soon? Would like to see it in production.

buunguyen:
Fixed in 2.0.8.

PR #245, opened 2016-01-25
Added a config for optionally path encoding (default true)

GodLesZ:
This implements a config for optional encoding the path component used in the `jsTree` elements.
I've got problems on my private GitLab installation with the encoded paths.

Possible that this could also be cached by changing some http routes or other server settings but I'm not the administrator so this was important for me.

Also works on Github and defaults to `true` to ensure the current behavior.

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -1,5 +1,6 @@
 class Adapter {
-  constructor(deps) {
+  constructor(store, deps) {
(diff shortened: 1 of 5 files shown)


buunguyen:
Why is this an option? Should we do it for all GitLab versions (and would it work)? Prefer users not have to think about this option as I guess most don't know what it is for and whether they need to turn it on or off.

PR #252, opened 2016-02-06
Use SVG icons for Github

crashbell:
I use `content` instead of `background` property to include SVG icons. This way has some weaknesses below: 
* Can't use CSS to change color, I had to add `style="fill: #acolor"` for specific icons. So the `gear` and `chevron` icons are unable to be highlighted when hovered.
* Can't resize the icon, the `branch` and `repo` are kinda big

We can work around with `background-image` but I had a trouble when applying `background-image` for pseudo `before` and `after`, just gave up. 

I placed the icons in outside of `src` folder since the `gulp` was broken when trying to read the SVG files. @buunguyen you may have a better place or some change for `gulp` script. 

I have tested with Safari and it works. No chance to test Firefox for now.


buunguyen:
Is there a reason you didn't base64 embed the icon to avoid a plethora of files?

buunguyen:
The two mentioned limitations seem quite bad. Probably need to investigate better solution.

joedf:
What about just injecting the icons/SVG in the HTML using JavaScript?

crashbell:
I added new code to solve 2 limitations but don't really like these tricky. Just let it here and we can fallback to it in case we can't find a better solution.

buunguyen:
@crashbell what less-loader? You meant gulp-less?

buunguyen:
@crashbell it's probably cleaner just embed svg directly in css so that we needn't a bunch of files and rely on loader / gulp-less. Something like this http://stackoverflow.com/a/10768631. What do you think?

joedf:
@buunguyen
I agree. :+1:
Actually, that is what @crashbell had in mind originally, not sure why he didn't do it that way...

crashbell:
@buunguyen @joedf you can see that they mentioned about `url-escaped`, I found out that the CSS file had a lot of SVG content which were escaped. It looked pretty dirty in code so I moved it to files to keep the CSS file clear. The loader will escape SVG content for us. Anyway, if you guys think that it's not necessary I will roll back when we want to stick with this solution.

joedf:
What about base64 encode the whole thing, like here?
http://stackoverflow.com/a/28866063/883015

buunguyen:
@crashbell when you have time, please reissue a PR with the 2 fixes. Also, please let me know which browsers you have tested.

PR #256, opened 2016-02-09
fixed: function getCreateTokenUrl is not defined

ruanyl:
When visiting repos which return `403 Forbidden`, for example: https://github.com/oh-my-fish/oh-my-fish
It shows `getCreateTokenUrl is not defined`

see the screenshot, in the second screenshot, octotree is empty.
5526a576-1d6f-4e1b-b7b9-cb42494906d4
32f2a99e-98a5-4e29-a055-088d8fc80a8e


buunguyen:
Thanks!

PR #260, opened 2016-02-11
fixed: popup overlay not removed when there is error

ruanyl:
This happens when octotree not able to access a repo(HTTP 403) example: https://github.com/oh-my-fish/oh-my-fish. But the 'empty' popup is not removed. which causes the link of `repo owner` and/or link to  the `repo` can not be reached.

See the screenshot below:
8d7dfe37-d7b7-42f4-9030-b651c91d90ef


buunguyen:
Thanks!

PR #267, opened 2016-02-24
Add ChromeDriver as an explicit devDependency

DrewML:
Test run after a fresh git clone + npm install fails attempting to run chromedriver.

```
Error: File does not exist: /Users/andrew/Documents/octotree/node_modules/chromedriver/lib/chromedriver/chromedriver
```


buunguyen:
Thanks a lot!

PR #268, opened 2016-02-25
Enable inline sourcemaps for octotree.js bundle

DrewML:
I started working on an adapter for BitBucket, and I added basic sourcemaps support for the main bundle to make debugging easier for myself.

I say basic because it's currently writing the sourcemaps inline, and there is no option to disable it. I think this has some benefits and some drawbacks.

Benefits:
- Easier debugging
- More accurate file/line # when users report issues w/ an exception from the console

Drawbacks:
- File size increase (`octotree.js` went from 56k >> 159k after making the change)

If you don't want to increase the size of the bundle for inline sourcemaps, we can write them externally and not include them in the final browser bundle(s). The only downside there is the loss of more accurate bug reports.

I didn't add sourcemaps for the content scripts (didn't think it would be super helpful). Having said that, let me know if you want me to change anything up.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -194,9 +194,11 @@ function buildJs(overrides, ctx) {
 
   return pipe(
     src,
+    $.sourcemaps.init(),
     $.babel(),
     $.concat('octotree.js'),
     $.preprocess({context: ctx}),
+    $.sourcemaps.write(),
     './tmp'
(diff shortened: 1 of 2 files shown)


buunguyen:
Probably use gulp-if to apply sourcemap conditionally. When `dist` runs, should disable.

Issue #186, opened 2015-09-11
Allow turning off download file by clicking icon


pdeva:
Since the latest update, whenever i click on the file icon it downloads it.
I really dont want that. Innumerous times this has resulted in me downloading a file when i was just clicking to see its contents in github.

This is very non-intuitive behavior since never does merely clicking an icon in tree item is supposed to download it to disk.
please atleast provide an option to disable it.

buunguyen:
Sorry, but I disagree that it's unintuitive given the icon is changed to the download icon upon mouse hover. The surface area of each tree item is large, you can click anywhere else to not trigger a download.

That said, I'm open to any suggestion to better implement the download feature.

halilim:
While not obscenely frequent, I, too, tend to hit the download icon inadvertently when quickly traversing a repo. Given that I rarely need to download a file, I'd love a harder-to-trigger implementation. 

Options

1. Place the download icon to the left of the file icon on hover (without changing the position of the file icon and the text).
2. Add a setting to disable the feature.

buunguyen:
@halilim sounds like a good approach. I'll accept a PR if anyone can work on it.

jamez14:
For anyone who wants to work on this, I did some work for disabling downloading altogether. But the implementation mentioned above is different than mine in #203

buunguyen:
This is a non-issue to me. If somebody wants it badly enough, please submit a PR.

jamez14:
@buunguyen what would your ideal spec be for the feature? What would it include (and not include)? I'd like to know just so if I work on this feature that I don't go against any of your plans/thoughts for this repo.

buunguyen:
@jamez14 sorry I thought I replied to this. I think an option should work (enabled by default).

PR #277, opened 2016-03-25
Enable/Disable downloading of a file via Settings

jamez14:
This change adds a setting to allow for the disabling of downloading a file. By default, the setting is checked (so downloading files will be enabled).

This change follows the same structure as the other settings, saving them in the store and retrieving them from the same place, by name. On hover, when disabled, the icon does not change and clicking the icon does nothing.

This PR should also satisfy the issue #186


src/view.tree.js
```diff
@@ -98,13 +106,18 @@ class TreeView {
...
 
     // handle icon click, fix #122
     if ($target.is('i.jstree-icon')) {
       $target = $target.parent()
-      download = true
+      if (enableDownloading) {
+        download = true
+      } else {
+        download = false
+      }
```
Ephemera: How about just `download  = enableDownloading`?

buunguyen:
I'm going to close this now. I don't think adding an extra option justifies the slight annoyance of having a download button.

Darkle:
I would really like to have this option. Quite often I accidentally click on the download icon by mistake when I am really trying to click on the file link to display the file as the space in between the icon and the filename will also cause the download to be triggered.

jibbers42:
My use case is a bit different. When looking at angular repos, I don't like the icon changing since angular has very long and similar file names. Often you can't see the extension without making the sidebar very wide. The icons are cool since you don't have to expand the sidebar; however, it's normal to mouse around what you're interested in, but doing so replaces the icon with a download icon :cry:

Ideally, I would prefer the idea at https://github.com/buunguyen/octotree/issues/186#issuecomment-167327341, but browsers make it very easy to right-click download the files, so I like this PR since it's easy and already done.

As a side note @buunguyen, it's a shame you approved the work in #186 and then decided to reject it once someone did it.

buunguyen:
Yeah, sorry about the change of mind. At first I thought it would be a simple change, but looking at the PR, it turns out to be quite complicated. Not to mention, an extra option users have to think about. I decided it's not worth it. One thing most people miss is that while it's easy for me to accept a PR, I am the one who has to maintain the code. Therefore, I need to balance between the value of a PR and the complexity it adds to the codebase.

PR #278, opened 2016-03-25
Fix middle click behaviors

### LINKED ISSUE #201: Support middle click properly
Love the extension, already insanely useful, good work!

I'm an extremely tab-heavy browser user, and tend to open links in background tabs for reading later once I've read a full page or when tracking interesting files in a GH repo. Currently, octotree seems to have a buggy middle-click implementation, a midlde-click will first open a new tab, as expected, but after a brief moment, the clicked link will be opened in the current tab as well. Expected behavior would be to not navigate anywhere in the current tab.

It might just be a `event.preventDefault()` missing somewhere, I haven't looked at the source at all so far.

itiut:
Fix middle click behaviors:
- Do not open links in the current tab. Fix #201.
- Directory links can be opened in new tab.

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -70,7 +70,9 @@ class Adapter {
             if (type === 'tree') {
               if (node) item.children = true
               else folders[item.path] = item.children = []
-              item.a_attr = { href: '#' }
+              item.a_attr = {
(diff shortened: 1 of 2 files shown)


buunguyen:
Released in 2.0.10. Thanks for the PR and sorry for the delayed merge.

PR #279, opened 2016-03-27
fix for showing above github fork overlay

kellyselden:
Fix for this:
capture

### DIFF
--- a/src/styles/base.less
+++ b/src/styles/base.less
@@ -10,7 +10,7 @@
   left: 0;
   bottom: 0;
   border-right: 1px solid #ddd;
-  z-index: 1000001;
+  z-index: 98; // github fork overlay is 99
   transition: transform .2s ease;
   transform: translate3d(-100%, 0, 0);
 


Mortiferr:
I've noticed this too. +1.

buunguyen:
Thanks for this. I'm not sure how this affects GitLab since the z-index style is shared. Can you check and confirm?

Mortiferr:
I don't use GitLab so @kellyselden can you take care of this?

buunguyen:
Thanks for checking. Yes, github and gitlab have separate style file (see the `adapters` directory). So you can try moving the z-index from base to individual style file.

kellyselden:
@buunguyen Will do soon.

Mortiferr:
But does GitLab even have something like the GitHub fork overly? I didn't notice the background going dark like that when I tried to fork something on GitLab.

Mortiferr:
Upon further investigation this isn't a problem at all on GitLab. This PR is good to go then as far as I can see. @buunguyen

buunguyen:
@Mortiferr GitLab has a sidebar that Octotree needs to show on top.

Issue #282, opened 2016-04-08
Support for self-hosted GitLab


ajsb85:
Can we add a new URL as option to use our instance of GitLab in these files: 
`background.js`
`manifest.json`

buunguyen:
You can follow instructions here https://github.com/buunguyen/octotree#enterprise-urls

ajsb85:
I try in different ways but I cannot see the permission prompt.
Can you make a screencast?
Thanks.

leocaseiro:
Hi @ajsb85, make sure you are not using Safari at it doesn't work there.
Try on Chrome.

This is a [step by step](https://github.com/buunguyen/octotree/issues/280#issuecomment-207054600) that might help you

Issue #280, opened 2016-04-04
gitlab could not use octotree


pippolover:
i use enterprise version of gitlab.
i could not find octotree button on new version of gitlab , could u have a look at this problem?
i try on gitlab.com and i could not find the button as well.

leocaseiro:
Hi @pippolover, you must add your gitlab url in settings from a gitlab.com repo first:

1. Go to https://gitlab.com/gitlab-org/gitlab-ce
2. Open octotree ![screen shot 2016-04-08 at 5 11 54 am](https://cloud.githubusercontent.com/assets/940070/14363599/76759d94-fd48-11e5-948b-319d8db60545.png)
3. Click at settings ![screen shot 2016-04-08 at 5 12 00 am](https://cloud.githubusercontent.com/assets/940070/14363614/8aded958-fd48-11e5-8d9d-eee7b1553093.png)
4. Than add your URL at **GitLab Enterprise URLs**
5. Click at **Save** button

After that, you're done :smile:

pippolover:
tks so much @leocaseiro .
problem fixed.

leocaseiro:
@buunguyen that would be nice add this step by step to `README.md`. What your thoughts?

buunguyen:
The readme does [mention it](https://github.com/buunguyen/octotree#enterprise-urls), but it might not be clear enough. If you think an improvement can be made, feel free to submit a PR.

PR #283, opened 2016-04-13
Add gitlab repo link

leocaseiro:
Add gitlab repo that might help some users as we discussed at #280 and #282


buunguyen:
Why does it have to be the GitLab CE repo?

leocaseiro:
Doesn't need to, but in my case, I don't have any GitLab repo and that was the less "branding" one I found so far. We can edit, something like:

> To add GitHub enterprise, go to a GitHub repo; otherwise, go to a GitLab repo (such as [gitlab-ce](https://gitlab.com/gitlab-org/gitlab-ce))

Is that nicer to you?

buunguyen:
Sorry for taking long to respond. Yes, that change would be great. Can you modify and submit?

leocaseiro:
Done!

buunguyen:
Thanks a lot!

PR #302, opened 2016-06-18
Update Regex for new GitLab layout

### LINKED ISSUE #301: After gitlab.com redesigned, extension works only in project tab
If you click on project it's working, in other cases it can't parse from pages(shows loader infinite loading :)).

crashbell:
Fix #301


buunguyen:
Thanks a lot, @crashbell!

PR #305, opened 2016-07-16
Update GitLab access token URL

josephfrazier:
It looks like this now:
https://gitlab.com/profile/personal_access_tokens

### DIFF
--- a/src/adapters/gitlab.js
+++ b/src/adapters/gitlab.js
@@ -74,7 +74,7 @@ class GitLab extends Adapter {
 
   // @override
   getCreateTokenUrl() {
-    return `${location.protocol}//${location.host}/profile/account`
+    return `${location.protocol}//${location.host}/profile/personal_access_tokens`
   }
 
   // @override


src/adapters/gitlab.js
```diff
@@ -74,7 +74,7 @@ class GitLab extends Adapter {
 
   // @override
   getCreateTokenUrl() {
-    return `${location.protocol}//${location.host}/profile/account`
+    return `${location.protocol}//${location.host}/profile/personal_access_tokens`
```
crashbell: Is that backward-compatible?

buunguyen:
Not sure if there's an easy way to address compatibility in this case. But I'd like to merge now so that it works with latest version.

josephfrazier:
Yeah... it looks like this won't work with GitLab < 8.8: https://github.com/gitlabhq/gitlabhq/blob/c6bfad2a672c4d2c623c8dd99a0dbc2138f325ab/doc/api/README.md#personal-access-tokens

Perhaps we could [check the GitLab version](https://stackoverflow.com/questions/21068773/checking-of-gitlab-version/22138457#22138457) somewhere and modify the URL as needed?

PR #308, opened 2016-07-18
Skip when a url ends with `.patch`, `.diff`

### LINKED ISSUE #307: Tree Overlaps content on patch files
In Github:

Example patch file: https://github.com/mitchellh/vagrant/commit/b57b0e0d48fe8b4196ca6b7e01bb6c1ecb4b69f9.patch

Notice the tree is overlapping the content. There probably are other pages that are like this as well, sounds like a simple blacklist addition?

Ephemera:
Fix #307

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -123,9 +123,9 @@ class GitHub extends Adapter {
       return cb()
     }
 
-    // (username)/(reponame)[/(type)]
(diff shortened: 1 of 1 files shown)


buunguyen:
@Ephemera this causes a bug https://github.com/buunguyen/octotree/issues/309, can you take a look?

Ephemera:
@buunguyen 
My mistake. I fixed with #310.

Issue #22, opened 2014-05-14
Support BitBucket


olimortimer:
Love the extension!

It would be great if it also displayed a tree for Bitbucket repos :)

maximkott:
+1

bimp:
+1

chodorowicz:
+1

leomperes:
+1

drkn:
+1

an-k:
+1

kxxoling:
+1

buunguyen:
I've just released Octotree 2.0 to Chrome store. This version supports GitLab using the same codebase. Hope the implementation of the GitLab adapter will inspire someone to implement an adapter for Bitbucket as I can't afford the time to do that myself. I will provide as much help as I can should anyone take on this work.

tkothe:
+1 for Bitbucket/Stash Support

jaxiinofea:
+1

buunguyen:
There's a WIP for BitBucket support here https://github.com/buunguyen/octotree/pull/328. Unfortunately, the original author @itiut hasn't responded to my requests for improvements. The PR looks pretty solid though and is pretty close to being complete. Anyone here wants to continue building upon this work?

mattblang:
@buunguyen I will try to take a look sometime over the holidays!

ans-4175:
any news again @mattblang @buunguyen or @itiut ?

tvytlx:
Recently, [Bitbucket Cloud gets a new navigation experience](https://blog.bitbucket.org/2017/04/11/bitbucket-cloud-gets-a-new-navigation-experience/), but it's still in beta.

To keep Octotree up to date, maybe we need to wait for a formal release of Bitbucket's UI before advancing this PR #328. (Actually, this PR works great in Bitbucket's old UI, why nobody finishes it...)

tvytlx:
Anyway, I finish it #407.

Maybe just waiting for a new release.

kasperite:
Hiya, any ETA for new release @buunguyen?

buunguyen:
Will release today.

minhkiller:
@buunguyen It is good to see you support bitbucket buddy.
I am finding some issues with the new Bitbucket UI

![image](https://user-images.githubusercontent.com/919899/29645620-1b46cb80-88c2-11e7-9719-875c90c35b65.png)

[...]

PR #328, opened 2016-09-22
[WIP] Add support for Bitbucket Cloud

itiut:
Add support for Bitbucket Cloud. Refs #22.
I confirmed that this works well on Chrome, Safari, and Firefox.

For now, this doesn't support for Bitbucket Server and may not work well for Bitbucket Cloud with Mercurial. (I'm sorry but I don't use Bitbucket Server and Mercurial.)

I have two concerns about implementation.

Firstly, there are [app passwords](https://blog.bitbucket.org/2016/06/06/app-passwords-bitbucket-cloud/) on Bitbucket, which are about the same functions as personal access tokens on GitHub, but app passwords can be used only for Basic Authentication. So the username is also required to access private repositories. Get the username embedded in HTML when the user is logged in, and use it for HTTP requests.

Secondly, git submodules can't be retrieved with Bitbucket API [src Resources](https://confluence.atlassian.com/bitbucket/src-resources-296095214.html) but can only by reading `.gitmodules` content. So when reading `.gitmodules` content for the first time, memoize it so that submodules will be inserted into the tree later.

And I have a question about CSS on Safari. Injected styles don't override Bitbucket original styles and the layout is skewed.

If there are something to be fixed, please point out them.

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -76,7 +76,7 @@ class Adapter {
               // encodes but retains the slashes, see #274
               const encodedPath = path.split('/').map(encodeURIComponent).join('/')
               item.a_attr = {
-                href: `/${repo.username}/${repo.reponame}/${type}/${repo.branch}/${encodedPath}`
+                href: this.getItemHref(repo, type, path)
               }
             }
             else if (type === 'commit') {
@@ -287,4 +287,12 @@ class Adapter {
   _getSubmodules(tree, opts, cb) {
     throw new Error('Not implemented')
   }
+
+  /**
+   * Returns item's href value.
+   * @api public
+   */
+   getItemHref(repo, type, encodedPath) {
+     return `/${repo.username}/${repo.reponame}/${type}/${repo.branch}/${encodedPath}`
+   }
 }
--- a/src/adapters/bitbucket.js
+++ b/src/adapters/bitbucket.js
@@ -1,3 +1,9 @@
+const BB_RESERVED_USER_NAMES = [
+  'account', 'dashboard', 'integrations', 'product',
+  'repo', 'snippets', 'support', 'whats-new'
+]
+const BB_RESERVED_REPO_NAMES = []
+
 class Bitbucket extends Adapter {
 
   constructor() {
@@ -45,6 +51,12 @@ class Bitbucket extends Adapter {
 
     const username = match[1]
     const reponame = match[2]
+
+    // Not a repository, skip
+    if (~BB_RESERVED_USER_NAMES.indexOf(username) ||
+        ~BB_RESERVED_REPO_NAMES.indexOf(reponame)) {
+      return cb()
+    }
 
     // TODO: get branch name
     const branch = ''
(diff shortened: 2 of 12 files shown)


buunguyen:
This is great, thank you! I'm a bit swamped right now, I'll provide feedback later.

src/adapters/adapter.js
```diff
@@ -287,4 +287,12 @@ class Adapter {
   _getSubmodules(tree, opts, cb) {
     throw new Error('Not implemented')
   }
+
+  /**
+   * Returns item's href value.
+   * @api public
```
buunguyen: should be @protected?

src/adapters/bitbucket.js
```diff
@@ -0,0 +1,255 @@
...
+        container: $pjaxContainer
+      })
+    }
+    else { // falls back
+      super.selectFile(path)
+    }
+  }
+
+  // @override
+  downloadFile(path, fileName) {
```
buunguyen: This looks like the implementation in base adapter, why redefining?

src/adapters/bitbucket.js
```diff
@@ -0,0 +1,255 @@
...
+    super(['jquery.pjax.js'])
+
+    $.pjax.defaults.timeout = 0 // no timeout
+    $(document)
+      .on('pjax:send', () => $(document).trigger(EVENT.REQ_START))
+      .on('pjax:end', () => $(document).trigger(EVENT.REQ_END))
+  }
+
+  // @override
+  init($sidebar) {
```
buunguyen: This method & selectFile() duplicate with github adapter a lot, maybe add a PjaxAdapter to keep the share code?

src/template.html
```diff
@@ -1,6 +1,6 @@
 <div>
   <nav class="octotree_sidebar">
-    <a class="octotree_toggle btn">
+    <a class="octotree_toggle btn aui-button">
```
buunguyen: This is too bb-specific, I'd prefer we style in bitbucket.less

src/adapters/bitbucket.js
```diff
@@ -0,0 +1,255 @@
...
+
+  _get(path, opts, cb) {
+    const host = location.protocol + '//' + 'api.bitbucket.org/1.0'
+    const url = `${host}/repositories/${opts.repo.username}/${opts.repo.reponame}${path || ''}`
+    const cfg  = { url, method: 'GET', cache: false }
+
+    if (opts.token) {
+      // Bitbucket App passwords can be used only for Basic Authentication.
+      // Get username of logged-in user.
+      const currentUser = JSON.parse($('body').attr('data-current-user'))
```
buunguyen: Maybe it's more reliable if we allow user to specify both username & password. We can use a single field a separate by `:`. Just change the verbiage in the case of BitBucket so that it's clear.

In other words, if the token has `:`, assume users enter both and use this username. If users don't explicitly specify username, retrieve it from the page (like currently). If can't retrieve (DOM changes), show error and ask user to enter both username & password, separated by `:`.

src/octotree.js
```diff
@@ -11,13 +11,26 @@ $(document).ready(() => {
   }
 
   function createAdapter() {
+    const normailzeUrl = (url) => url.replace(/(.*?:\/\/[^/]+)(.*)/, '$1')
```
buunguyen: typo, normalizeUrl

buunguyen:
@itiut did you have a chance to continue working on this?

PR #332, opened 2016-10-04
open files in a new tab if ctrl is pressed

fieteboerner:
opens files, submodules and repositories in a new tab by clicking a link
while the ctrl button is pressed

this functions allows the user to serve multiple files simultaneously
and it provides the common feeling of an a-tag

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -252,11 +252,29 @@ class Adapter {
   }
 
   /**
+   * Selects the file at a specific path in a new tab window.
+   * @api public
+   */
+  selectFileNewTab(path) {
+    let win = window.open(path, '_blank')
+    win.focus()
+  }
+
+  /**
    * Selects a submodule.
    * @api public
    */
   selectSubmodule(path) {
     window.location.href = path
+  }
+
+  /**
+   * Selects a submodule in a new tab window.
+   * @api public
+   */
+  selectSubmoduleNewTab(path) {
+    let win = window.open(path, '_blank')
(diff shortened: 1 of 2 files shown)


src/adapters/adapter.js
```diff
@@ -252,6 +252,15 @@ class Adapter {
   }
 
   /**
+   * Selects the file at a specific path in a new tab window.
+   * @api public
+   */
+  selectFileNewTab(path) {
```
buunguyen: These 2 methods do the same thing. Maybe rename to a single one: `openInNewTab(path)`.

src/view.tree.js
```diff
@@ -64,7 +64,7 @@ class TreeView {
       )
       .on('click', 'a[data-pjax]', function (event) {
         event.preventDefault()
-        adapter.selectFile($(this).attr('href') /* a.href always return absolute URL, don't want that */)
+        event.ctrlKey ? adapter.selectFileNewTab($(this).attr('href')) : adapter.selectFile($(this).attr('href')) /* a.href always return absolute URL, don't want that */
```
buunguyen: Too long, should be 2 line, one to get the href and another to select it.

phukiensongphat:
Hi all, anybody can to tell me where i can learn php , because i have problem with my `website[http://phukiensongphat.com/chitietsanpham/may-nghe-phat-phap-tung-kinh-giang-kinh-niem-phat-50.html](url) can not share to facebook, if i copy link to this website, i can not see images, anybody help me, thanks a lot

PR #337, opened 2016-10-19
Hide octotree in print mode

broven:


broven:
one
two

src/styles/base.less
```diff
@@ -279,3 +279,11 @@ a.octotree_toggle {
     }
   }
 }
+@media print {
```
buunguyen: Thanks, can you reformat this a bit:

* Add new line
* Space before `{`

PR #352, opened 2016-12-11
339 ff download

scilganon:
regarding to:
 - https://support.mozilla.org/en-US/questions/968992
 - http://stackoverflow.com/questions/27115748/download-attribute-not-working-in-firefox/27116581#27116581

need testing

### DIFF
--- a/src/config/firefox/overrides.js
+++ b/src/config/firefox/overrides.js
@@ -1,0 +1,10 @@
+(() => {
+  Adapter.prototype.downloadFile = function(path, fileName){
+    const link = document.createElement('a')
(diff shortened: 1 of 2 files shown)


buunguyen:
Thanks for this. Does this work with Chrome and Safari too? If it does, I suggest changing the code in adapter directly instead of creating an override.

src/config/firefox/overrides.js
```diff
@@ -0,0 +1,10 @@
+(() => {
+  Adapter.prototype.downloadFile = function(path, fileName){
+    const link = document.createElement('a')
+    link.setAttribute('href', 'data:text/plain;charset=utf-8,' + path.replace(/\/blob\//, '/raw/'))
```
buunguyen: Would this work with non-text file?

PR #367, opened 2017-02-11
Adapt styles to new github header design

dimorinny:
Now octotree adapted for old github design. In this pull request I changed styles, some icons and now it's look like native github header. Also I changed screenshots in docs.


crashbell:
@dimorinny Can you please resize the screenshots?

dimorinny:
@crashbell No problems, but for what size I should resize that?

crashbell:
@dimorinny How about current size of [chrome-github.png](https://github.com/buunguyen/octotree/blob/master/docs/chrome-github.png) W: 1800px | H: 1179px ?

buunguyen:
Thanks for this @dimorinny. 

I've been sitting on the fence because I find the new GitHub design a bit weird. Changing Octotree header to that design looks even weirder as the contrast is a bit off. Maybe we should only change the height of Octotree header and keep the current color? Or maybe changing Octotree panel/body so that it looks more harmonious with the dark header?

Love to hear what everyone thinks.

dimorinny:
@buunguyen I agree with you, that new header is controversial decision. But I think that octotree should firstly looks harmonious with service, where it inserted.

dawikur:
I think fixing height is a must.
Now, when I am looking at this I see that github changed also color of the grey-ish strip under the header, it is visible here:
mixed_colors

The new one colder and probably better fits new header. We should probably change also this one.

mateusmarquezini:
Hey, folks

I didn't understand the changes about this issue. By the way, can I change the current header design or not so far?

Thanks.

PR #374, opened 2017-02-17
WIP: remove turbolinks from GitLab, it is no longer used there and breaks …

### LINKED ISSUE #373: GitLab removed turbolinks
### Description

GitLab recently removed turbolinks, now navigating between pages causes a bunch of stuff to not load. 

[GitLab CE issue](https://gitlab.com/gitlab-org/gitlab-ce/issues/28161)

### Environment (if bug)

* Octotree version: 2.0.20
* Browser & version: Chrome 56
* OS & version: All

psimyn:
…things

### Description

Fixes #373 - Turbolinks on GitLab prevents JS from loading when navigating between pages. Maybe this is fixable and keeping TurboLinks but this seemed like the easier option

### Proposal

Removed refs to turbolinks I could find

### Todos

- [ ] Don't reanimate opening when navigating between links


crashbell:
Thanks @psimyn. Disabling Turbolinks is the right thing we need to do but we also need an alternative way to detect Ajax reload. With your fix, Octotree sidebar will reload and animate whenever the links change. I'm exploring GitLab to see what they use for Ajax reload. Hopefully we can detect that easily.

buunguyen:
Does ajax loading still work after removal?

psimyn:
removes ajax loading. Noticable thing for _most_ pages is that the sidebar reanimates on page loads, so this will need a change to init as expanded when it is first rendered.

As mentioned on the issue, there are still a few places where GitLab uses ajax-loaded pages that  we should handle. I can have a better look at this over the next day or two, otherwise @crashbell might have some more findings.

buunguyen:
I will remove GitLab support.

PR #391, opened 2017-03-21
Fixed typo.

chrisbrownie:
This PR resolves a typo displayed when private repos are connected to without an access token:

**Before:**
```Accessing private repositories requires a access token.```

**After:**
```Accessing private repositories requires an access token.```


buunguyen:
Thanks!

Issue #157, opened 2015-01-25
Octotree browses master when page browses commit


Topher-the-Geek:
When browsing a branch (or tag), Octotree follows along. For example, if I use the branches/tags dropdown on the GitHub page to choose a branch (or tag), Octotree does the right thing.

![branch](https://cloud.githubusercontent.com/assets/588423/5893138/84676eec-a48d-11e4-9385-e58978f3a92f.jpg)

However, when browsing a commit, Octotree shows master. For example, if I click "Browse Tree" when looking at a CL in the history, Octotree shows master rather than the commit. Furthermore, clicking on directories and files in Octotree switches the URL to the master branch rather than the commit hash.

![commit-hash](https://cloud.githubusercontent.com/assets/588423/5893146/bdaa5a2a-a48d-11e4-848f-296a8db312d9.jpg)

This is an awesome plugin! Small bugs aside.

buunguyen:
Supported in 1.6 (releasing soon).

PR #398, opened 2017-04-07
New fallback to extract branch/commit on github

veelenga:
#157 seems to be broken or not fixed. Here is my proposal.

Adds a new fallback to extract branch/commit from page title. Result:

screen shot 2017-04-07 at 1 35 34 pm

Examples of titles that match:

<pre>
  * buunguyen/octotree at 2b63178b6e5072bf358797d25913fdc6e0d51210
  * octotree/test at 2b63178b6e5072bf358797d25913fdc6e0d51210 · buunguyen/octotree
  * octotree/docs at master · buunguyen/octotree
</pre>

Examples of titles that do not match:

<pre>
  * buunguyen/octotree: Code tree for GitHub
  * Fixed typo. (#391) · buunguyen/octotree@8833647
</pre>

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -155,6 +155,8 @@ class GitHub extends Adapter {
     const branch =
       // Code page
       $('.branch-select-menu .select-menu-item.selected').data('name') ||
+      // Page title
+      (document.title.match(/ at (\w+)/) || '')[1] ||
       // Pull requests page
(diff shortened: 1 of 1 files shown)


veelenga:
Hm, looking again into this, i understood that this will not work correctly in all cases. Here is an example of title, where branch would be wrongly matched:

<pre>
Fixed typo at src/main.js · buunguyen/octotree@8833647
</pre>

Need to find better approach to fix this. Sorry

PR #397, opened 2017-04-05
Filter tree based on PR

jefftougas:
### Description
This addresses #313 - I took inspiration from @mcuadros's [fork](https://github.com/buunguyen/octotree/compare/master...mcuadros:master), but it wasn't working for me so I built my own.

### Proposal
- On PR file list / diff tab in PRs, filter the tree to only show files / folders that have a diff in the PR
- In PR view mode, clicking any file in the tree takes you to that file's diff within the PR
- Diff stats are shown next to each folder / file in the tree

image

PR #414, opened 2017-07-13
Add jstree search feature

maliming:
### Add jstree search feature
fast location files, support case sensitive, fuzzy search, etc. jstree search option.

About [jstree search plugin:](https://www.jstree.com/api/#/?q=$.jstree.defaults.search&f=$.jstree.defaults.search)

### DIFF
--- a/libs/bindWithDelay.js
+++ b/libs/bindWithDelay.js
@@ -1,0 +1,53 @@
+/*
+bindWithDelay jQuery plugin
+Author: Brian Grinstead
+MIT license: http://www.opensource.org/licenses/mit-license.php
+http://github.com/bgrins/bindWithDelay
+http://briangrinstead.com/files/bindWithDelay
+Usage:
+    See http://api.jquery.com/bind/
+    .bindWithDelay( eventType, [ eventData ], handler(eventObject), timeout, throttle )
+Examples:
+    $("#foo").bindWithDelay("click", function(e) { }, 100);
+    $(window).bindWithDelay("resize", { optional: "eventData" }, callback, 1000);
+    $(window).bindWithDelay("resize", callback, 1000, true);
+*/
+
+(function($) {
+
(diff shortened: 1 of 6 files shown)


crashbell:
Thanks for your work. That basically looks good to me. I also have some comments below:
- The search works locally, that means, you can't search the files haven't been loaded. That seems not very useful in lazy-mode. That would be great if we can search the whole repo in lazy-mode (I do know it's tough, especially, for large repos).
- The link of search options is not intuitive. Should we write our own description as [Settings](https://github.com/buunguyen/octotree#settings)?
- The search options naming should be consistent with existing options.

maliming:
@crashbell Very good advice, I also thought about the relevant issues before, do not know @buunguyen  is willing to join the search feature. If he is willing, I will continue to improve this feature.

buunguyen:
This looks good. Only thing is people don't search all the times yet the search box appears in quite valued estate on the top. Maybe the search box is only displayed when people explicitly click a search icon or hit a short key combination? Also, I think the styling of the matched items can be improved, bold & italic don't look good IMO.

Agreed with the points @crashbell mentioned, esp. the point about simplifying the search options. The 1st could be tough but may not be of very high priority. If people click on nodes to expand in lazy mode and the matching still works with newly added items, I think that's probably fine.

maliming:
I will continue to improve, hope that the next PR can let everyone accept, Thank you @crashbell @buunguyen the proposal that is very helpful to me.

marsjaninzmarsa:
Maybe just leave this PR open, mark it as [WIP] and continue to working on it. Open PR → Better visibility → Chances for helpful colab and minimize risk of work duplication.

marsjaninzmarsa:
Or just claim #397 as current WIP for this, that's confusing.

buunguyen:
@maliming are you still interested in working on this?

rajendraarora16:
Looking forward

Jinkwon:
Looking forward so much 👍

PR #428, opened 2017-08-23
Filter tree based on Pull Request (Github only)

bigredwill:
### Description
This PR addresses https://github.com/buunguyen/octotree/issues/313.  These changes are a fork of @jefftougas' fork, which was submitted as PR https://github.com/buunguyen/octotree/pull/397.

This PR is different in that it uses the Github API to retrieve the PR diff from the files changed endpoint: https://developer.github.com/v3/pulls/#list-pull-requests-files

It also adds an option to the settings for `Show only changed files in a pull request`.
### Proposal
* On PR file list / diff tab in PRs, filter the tree to only show files / folders that have a diff in the PR
* Diff stats are shown next to each folder / file in the tree
* Only filter tree if option for `Show only changed files in a pull request` is true in settings.
* In PR view mode, clicking any file in the tree takes you to that file's diff within the PR
  * NOTE: This was not completed. Clicking a file has the same result as clicking a file in the non filtered view.  As far as I can tell, there is no way to retrieve a direct link to a file on the `Files Changed` page in a PR.  It is possible in the other PR because the page itself is being parsed.  If anyone knows a way to retrieve this without parsing the DOM on the `Files Changed` page, please let me know.

#### Settings Screen
screen shot 2017-08-22 at 5 18 28 pm

#### Unexpanded Filtered Tree
screen shot 2017-08-22 at 3 29 30 pm

#### Expanded Filtered Tree
screen shot 2017-08-22 at 3 30 04 pm


src/adapters/adapter.js
```diff
@@ -60,6 +60,35 @@ class Adapter {
...
+                 else if (item.patch.action === 'rename') {
+                   patch_html +=
+                   `<span class="text-green"
+                     title="${item.patch.previous}">renamed
+                    </span>`
+                 }
+               }
+
+               if (item.patch.changes) {
+                 patch_html += `<span>${item.patch.changes} changes</span>`
```
jeremyalan: I would consider changing the text to `# files` instead of `# changes` -- it's a bit misleading as-is, as I would expect changes to total the number of lines that were modified.

src/adapters/github.js
```diff
@@ -149,10 +153,93 @@ class GitHub extends PjaxAdapter {
...
+              path = `${path}/${curr}`
+            }
+            else {
+              path = `${curr}`
+            }
+            // Path already has been recorded, accumulate changes
+            if (typeof diffMap[path] === 'object') {
+              diffMap[path].additions += file.additions
+              diffMap[path].deletions += file.deletions
+              diffMap[path].changes++
```
jeremyalan: Related to my previous comment, consider changing this property to `files` or similar.

src/adapters/github.js
```diff
@@ -149,10 +153,93 @@ class GitHub extends PjaxAdapter {
...
+            }
+            // Path is new
+            else {
+              diffMap[path] = {
+                additions: file.additions,
+                deletions: file.deletions,
+                changes: 1,
```
jeremyalan: Related to my previous comment, consider changing this property to `files` or similar.

bigredwill:
@jeremyalan done! This has been too long in the pipeline, I can't wait either 👍

bigredwill:
#### Updated UI, `files` instead of `changes` 
screen shot 2017-08-22 at 9 20 26 pm

buunguyen:
@crashbell would love to hear what you think about this PR.

jchannon:
This is awesome!!🎉🎉

buunguyen:
@bigredwill great stuff, thanks for working on this. 

I've tried it out and here's some issues:

1. It doesn't work when the option "Load entire tree at once" is unchecked (i.e. lazy loading mode). I'm not sure if it's feasible to make this feature works with lazy loading mode; if not, at least we should probably disable this option unless "Load entire tree at once" is enabled.

2. (minor thing) The verbiage of "Show only changed files..." is a bit long, it wraps for default sidebar width. Suggest rephrase to simply "Show only changes in pull requests".

3. When clicking on a file in the PR view, I think it's better that it links to the PR diff, like https://github.com/buunguyen/octotree/pull/428/files#diff-ab93c06aa4a7b7156444a9205bb16c21. For now, clicking will jump to the code page, which has 2 effects: the file doesn't contain the change and it reloads the tree (i.e. full tree instead of PR tree).

What do you think?

timglabisch:
thanks, awesome. 


>  When clicking on a file in the PR view, I think it's better that it links to the PR diff,

👍 👍 👍

bigredwill:
Re:
1.  I'll look in to this, either making it work with lazy loading mode or disabling if not feasible.
2.  Will update.
3.  This is an issue in that the hash/id appended to the url to link to a specific file is not actually the `sha` of the file diff.  I'm not sure where that hash/id is generated, and it is not available through any Github API as far as I can tell.  
That said, this is doable.  I inspected the DOM structure of the `Files Changed` tab, and each file has another id associated with it that is a 0 indexed number and assigned in alphabetical order.  So the first file in that tab is `#diff-0`, the second is `#diff-1`, etc.  To allow for jumping to the `Files Changed` tab in the correct location, I'll add an `diffId` field to the file in `_getPatch` that allows us to generate a correct href.

timglabisch:
@bigredwill why not just jumping to the element with the data-path?

data_path

bigredwill:
@timglabisch I don't believe you can use the url to scroll to an element based on a `data-` attribute.    

The hash/id appended to the url tells the browser to jump to the first element on the page with that hash/id as it's `id` or `name` attribute.

See the image below.  The `div` container for `src/adapters/bitbucket.js` contains the attribute `id="data-1"`.  This is because it is the second file changed on the page.

screen shot 2017-08-23 at 2 02 17 pm

bigredwill:
@buunguyen These three recent commits should take care of all the issues you presented :)

Works with lazy loading as well 🎉

src/adapters/github.js
```diff
@@ -149,10 +153,98 @@ class GitHub extends PjaxAdapter {
...
+        else {
+          this._getPatch(opts, (patchErr, patchRes) => {
+            const diffExists = patchRes && Object.keys(patchRes).length > 0
+            if (patchErr || !diffExists) cb(null, res.tree)
+            else {
+              // Filter tree to only include files and directories that are included in the patch
+              const filteredTree = res.tree
+                  .filter((node) => {
+                    // If lazy load, prepend the path of node that's currently loading
+                    let nodePath = opts.node.path ? `${opts.node.path}/${node.path}` : node.path
```
crashbell: I don't see we have a null-check here for `opts.node`. Can it cause an error?

src/adapters/github.js
```diff
@@ -149,10 +153,98 @@ class GitHub extends PjaxAdapter {
...
+                    return patchRes[nodePath] !== undefined
```
buunguyen: We probably don't need the `map` call by adding:

```js
node.patch = patchRes[nodePath]
return !!node.patch
```

src/adapters/github.js
```diff
@@ -149,10 +153,98 @@ class GitHub extends PjaxAdapter {
...
+          // Start at root folder and construct path
+          split.reduce((path, curr) => {
+            if (path.length) {
+              path = `${path}/${curr}`
+            }
+            else {
+              path = `${curr}`
+            }
+            // Path already has been recorded, accumulate changes
+            if (typeof diffMap[path] === 'object') {
```
crashbell: I think we can make it shorter with `if (diffMap[path])`.

src/adapters/adapter.js
```diff
@@ -60,6 +60,36 @@ class Adapter {
...
+                    </span>`
+                }
+              }
+
+              if (item.patch.filesChanged) {
+                const fileString = item.patch.filesChanged === 1 ? 'file' : 'files'
+                patch_html += `<span>${item.patch.filesChanged} ${fileString}</span>`
+              }
+
+              if (item.patch.additions !== 0 || item.patch.deletions !== 0) {
```
crashbell: For zero number, I think it makes more sense to have it empty. `+0` or `-0` is not informative to me.

src/adapters/adapter.js
```diff
@@ -73,10 +103,17 @@ class Adapter {
                 else folders[item.path] = item.children = []
               }
 
-              // encodes but retains the slashes, see #274
-              const encodedPath = path.split('/').map(encodeURIComponent).join('/')
-              item.a_attr = {
-                href: this._getItemHref(repo, type, path)
+              // If item is part of a PR, jump to that file's diff
+              if (item.patch && typeof item.patch.diffId === 'number') {
```
crashbell: `typeof item.patch.diffId === 'number'` => I'm a bit curious whether there're other data types.

crashbell:
Thanks @bigredwill I have added some comments.

src/adapters/adapter.js
```diff
@@ -60,6 +60,36 @@ class Adapter {
             item.text = name
             item.icon = type // uses `type` as class name for tree node
 
+            if (item.patch) {
+              let patch_html = ''
+
+              if (item.patch.action) {
+                if (item.patch.action === 'add') {
```
buunguyen: Also, I don't think the outer `if (item.patch.action)` is necessary.

src/adapters/adapter.js
bigredwill: Initially I was checking `if (item.patch.diffId)`, but that returns false when `diffId === 0`

src/adapters/adapter.js
```diff
@@ -73,10 +103,17 @@ class Adapter {
...
-              item.a_attr = {
-                href: this._getItemHref(repo, type, path)
+              // If item is part of a PR, jump to that file's diff
+              if (item.patch && typeof item.patch.diffId === 'number') {
+                item.a_attr = {
+                  href: `/${repo.username}/${repo.reponame}/pull/${repo.pull}/files#diff-${item.patch.diffId}`
+                }
+              } else {
+                // encodes but retains the slashes, see #274
+                const encodedPath = path.split('/').map(encodeURIComponent).join('/')
```
bigredwill: I noticed the `encodedPath` thing as well, want me to update that with this PR?

src/adapters/github.js
```diff
@@ -85,23 +85,28 @@ class GitHub extends PjaxAdapter {
...
+    const match = window.location.pathname.match(/([^\/]+)\/([^\/]+)(?:\/([^\/]+))?(?:\/([^\/]+))?/)
     if (!match) {
       return cb()
     }
 
-    const username = match[1]
-    const reponame = match[2]
+    let username = match[1]
+    let reponame = match[2]
+    let type = match[3]
```
bigredwill: I think that makes sense

src/adapters/adapter.js
buunguyen: Yes, please. I think you'll need to change the case for patch too, if username/repo has forbidden characters which need encoding.

src/adapters/github.js
```diff
@@ -149,10 +153,98 @@ class GitHub extends PjaxAdapter {
...
-      else cb(null, res.tree)
+      else {
+        if (!opts.repo.pull) cb(null, res.tree)
+        else {
+          this._getPatch(opts, (patchErr, patchRes) => {
+            const diffExists = patchRes && Object.keys(patchRes).length > 0
+            if (patchErr || !diffExists) cb(null, res.tree)
+            else {
+              // Filter tree to only include files and directories that are included in the patch
+              const filteredTree = res.tree
```
bigredwill: Another bug I just caught with this code:
If a file is added or renamed, it doesn't exist in `res.tree`.  This means these files won't show up in the tree view.

To fix this, we will only use the result from `_getPatch` and return an array with objects that are similarly structured to those returned by the `/git/trees/${path}` call.  This has the added benefit of only making one network call rather than two as it is now.

buunguyen: @bigredwill sounds good. In that case, I suppose when PR mode is used, the lazy mode is not used. I think that's even better, avoid one source of bug and it isn't necessary to lazy load PR changes as lazy load only makes sense for huge repo, not a single change set.

buunguyen:
@bigredwill thanks for the update. 

I've just noticed file download doesn't work in PR mode:
* The file name is wrong (there's the "added", "renamed" suffix).
* The content is not the file itself, but the entire HTML page to the diff link.

Can you take a look?

Other than that, seems great, I'm very keen on merging this PR.

bigredwill:
@buunguyen still working on making this work and cleaning up the code.  Also updating per @crashbell 's suggestions.

About the file download.  Do we want to download the file diff? Or the full file itself?
See what's available in https://developer.github.com/v3/pulls/#list-pull-requests-files

buunguyen:
I think it should download the entire file (with the changes). If people want to see the diff, the can already click and see it here.

bigredwill:
@buunguyen These latest changes should fix the downloading.  

I had to add `data-downloadHref` and `data-downloadFileName` attributes so that we could access these from within the click handler.  Because the link in the PR view is not the actual file blob page but to the diff on the Files Changed page, these two attributes are needed.  The `downloadHref` attribute gives us the link to the blob, and the `downloadFileName` attribute gives us the name of the file (rather than rely on the anchor element's inner text).

src/adapters/github.js
```diff
@@ -147,11 +152,104 @@ class GitHub extends PjaxAdapter {
...
+      this._getPatch(opts, (err, res) => {
+        if (err) cb(err)
+        else cb(null, res)
+      })
+    }
+    else {
+      this._get(`/git/trees/${path}`, opts, (err, res) => {
+        if (err) cb(err)
+        else {
+          console.log(res.tree)
```
crashbell: Is this redundant?

src/adapters/github.js
```diff
@@ -147,11 +152,104 @@ class GitHub extends PjaxAdapter {
...
+    const {pullNumber} = opts.repo
+    this._get(`/pulls/${pullNumber}/files`, opts, (err, res) => {
       if (err) cb(err)
-      else cb(null, res.tree)
+      else {
+        const diffMap = {}
+        // Iterate files/folders to determine diff details
+        res.forEach((file, index) => {
+          // Grab parent folder path
+          const folderPath = file.filename.split('/').slice(0,-1).join('/')
```
crashbell: Please add a space between arguments: `slice(0, -1)`

src/adapters/github.js
```diff
@@ -147,11 +152,104 @@ class GitHub extends PjaxAdapter {
...
+          diffTree.push({
+            patch,
+            path: fileName,
+            sha: patch.sha,
+            type: patch.type,
+            url: patch.blob_url,
+          })
+        })
+        // Sort by path, needs to be alphabetical order (so parent folders come before children)
+        diffTree.sort((a,b) => a.path.localeCompare(b.path))
```
crashbell: `(a,b)` => `(a, b)`

src/adapters/github.js
bigredwill: Yup, was debugging, need to remove this..

bigredwill:
@crashbell most recent commits clean up code.
@buunguyen 
I fixed the case for `removed` patch status.  It looks like this now:
screen shot 2017-08-29 at 11 12 55 am

buunguyen:
@bigredwill awesome, thanks!

PR #433, opened 2017-08-29
Raise pr-tree response limit to max (300)

bigredwill:
### Description
The call made to get the files for a specific pull request defaults to only 30 per response. Larger PR's can have more than that many files.

### Proposal
The max that can be returned is 300 (per https://developer.github.com/v3/pulls/#list-pull-requests-files), so raised the limit to that.

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -180,7 +180,7 @@ class GitHub extends PjaxAdapter {
    */
    _getPatch(opts, cb) {
     const {pullNumber} = opts.repo
-    this._get(`/pulls/${pullNumber}/files`, opts, (err, res) => {
+    this._get(`/pulls/${pullNumber}/files?per_page=300`, opts, (err, res) => {
       if (err) cb(err)
       else {
         const diffMap = {}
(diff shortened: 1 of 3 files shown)


bigredwill:
@buunguyen ^this might be pretty important to be released with https://github.com/buunguyen/octotree/pull/428

We also might want to add a disclaimer that it will only show a maximum of 300 files. Perhaps this would be best in the settings view. Thoughts?

buunguyen:
Agreed, maybe show it below the option?

buunguyen:
How about?

`Note: allows maximum of 300 files`

bigredwill:
screen shot 2017-08-29 at 12 52 43 pm

bigredwill:
`allows` is a little ambiguous.

How about: `Note: displays maximum of 300 files`

buunguyen:
Great, that's better.

bigredwill:
Done. @buunguyen we'll probably want to add a description about this new setting in the README

buunguyen:
Thanks! I will update the readme.

PR #435, opened 2017-08-30
Fix issue with lazy loading and PR view

bigredwill:
### Description
This PR fixes issue https://github.com/buunguyen/octotree/issues/434, where the combination of lazy loading and pull request view was causing recursive trees to appear.

To reproduce the issue, navigate to any PR and change your settings to:
`Load entire tree at once: false`
`Show only pull request changes: true`

### Proposal
To fix, we add logic to check if we are in Pull Request mode to `view.tree.js`

Here is the tree in the repo linked to in https://github.com/buunguyen/octotree/issues/434 with the fix: 
screen shot 2017-08-30 at 4 53 55 pm

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -23,7 +23,9 @@ class TreeView {
     $jstree.settings.core.data = (node, cb) => {
       const loadAll = this.adapter.canLoadEntireTree() &&
                       this.store.get(STORE.LOADALL)
-      node = !loadAll && (node.id === '#' ? {path: ''} : node.original)
+      // No lazy loading in PR tree view
+      const isPR = this.store.get(STORE.PR) && repo.pullNumber
+      node = !loadAll && !isPR && (node.id === '#' ? {path: ''} : node.original)
 
       this.adapter.loadCodeTree({repo, token, node}, (err, treeData) => {
         if (err) {
@@ -31,7 +33,7 @@ class TreeView {
         }
         else {
           treeData = this._sort(treeData)
-          if (loadAll) {
+          if (loadAll || isPR) {
             treeData = this._collapse(treeData)
           }
           cb(treeData)


src/view.tree.js
```diff
@@ -23,15 +23,17 @@ class TreeView {
     $jstree.settings.core.data = (node, cb) => {
       const loadAll = this.adapter.canLoadEntireTree() &&
                       this.store.get(STORE.LOADALL)
-      node = !loadAll && (node.id === '#' ? {path: ''} : node.original)
+      // No lazy loading in PR tree view
+      const isPR = this.store.get(STORE.PR) && repo.pullNumber
```
buunguyen: This seems to leak GitHub specific logic into the UI.  I suggest encapsulate it into `adapter#canLoadEntireTree` if possible.

buunguyen: Actually, that's a bad idea :D. Let me think more about this.

src/view.tree.js
```diff
@@ -23,15 +23,17 @@ class TreeView {
...
+      node = !loadAll && !isPR && (node.id === '#' ? {path: ''} : node.original)
```
buunguyen: I'm not clear why this is necessary and whether we can simplify or get rid of this?

src/view.tree.js
bigredwill: This is needed because we need to call `this._collapse` down on line 34 to collapse nested empty folders. Without the added `|| isPR` check, the tree looks like this:
screen shot 2017-08-31 at 2 27 13 pm

buunguyen: I got that part, I was asking about line 28 `node = ...`.

src/view.tree.js
bigredwill: This is necessary because `loadCodeTree` expects `node` to be null if loading the entire tree.  When it constructs the tree view in `_loadCodeTreeInternal` it uses `node` to determine how to display the tree. When `node` is not null, it causes the "_recursive_"  bug that the issue outline.  

We can move that `isPR` check to that function in the adapter. The issue is we still need to check if `isPR` on line 36 so that we can collapse the tree.

buunguyen: Ok, thanks. I missed that it was the original logic and you just add the !isPR check to that. Been a while since I looked at this area of the code. I'm gonna merge this fix and see if I can refactor somehow.

buunguyen: I refactored a bit https://github.com/buunguyen/octotree/commit/879a2bbd3117f4655118ff66ec1dbae3c8155c7e. Thanks again for working on this!

PR #439, opened 2017-09-12
background: fix error while running permissions.remove

Ephemera:
### Description
bitbucket url should be added to whitelist because it is required permissions.
you can see the error in extension developer tools.


buunguyen:
Thanks!

PR #444, opened 2017-10-04
Skip pjax loads if requested path is for different anchor on same page

kevinconaway:
### Description
@buunguyen, when reviewing files on Github, each page click issues a pjax request for every file clicked on.  This seems unnecessary to me since all files are on the same page just with different anchors.  It would be better to just navigate to the requested anchor instead of refetching the page.

### Proposal
This modifies `pjax.selectFile` to skip the fetch if the request is for a different anchor on the same page that the user is already on


buunguyen:
Great, thanks!

Issue #445, opened 2017-10-06
Buttons crush the title


ue:
### Description
Setting button and open/close button it`s crush the title

### Environment (if bug)

* Octotree version: 2.3.1
* Browser & version: 61.0.3163.100 latest version
* OS & version:
* Screenshot, if any (drag an image here)

![image](https://user-images.githubusercontent.com/9980498/31266294-c552e930-aa7b-11e7-9469-0e4bb4e48664.png)

FlyteWizard:
Was just going to open an issue for this 👍 

Would be nice to have it not overlap.

buunguyen:
This is no longer an issue.

PR #446, opened 2017-10-06
new position for buttons and fixed issue #445

ue:
### Description
This is a quick fix.
I have change top pixels for design end crusning.
Could you please take a look the pictures. 
And it`s look like be cool I think :)

### Proposal

image
image

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -144,7 +144,7 @@
   }
 
   a.octotree_opts {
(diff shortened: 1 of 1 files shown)


buunguyen:
I think a better fix is reduce the width of the text. The button should align vertical as currently.

ue:
@buunguyen Alright I gonna work for that

PR #448, opened 2017-10-11
Branch switcher in header

ChaseWagoner:
### Description
Add a branch switcher to the header, if the repository has multiple branches. This addresses issue https://github.com/buunguyen/octotree/issues/317.

### Proposal
Allow users to switch between branches of the repository.

From the linked issue:
>Selecting a branch would load the related tree but not navigate until an item in the tree is clicked.

Instead of following this suggestion, this code will navigate immediately once a branch is selected, mimicking the behavior of GitHub's built-in branch dropdown.

### Demo

I created these branches in my fork of this repo:
- `master`
- `branch-switcher`
- `nonsense<branch&another`

I've since added branches with more 'special' characters ([my fork](https://github.com/ChaseWagoner/octotree)).

Here's a basic demo:

branch-switcher

### Discussion (non-exhaustive list)
- I disabled switching to a branch when its `deXss`ed name doesn't match its raw name, because I don't know the complete list of characters GitHub allows/encodes in URLs, and I don't want to guess incorrectly.
- This doesn't work well when a file doesn't exist.
    - Example: [this file](https://github.com/ChaseWagoner/octotree/blob/temp/branch/extra-file.txt) exists only on one branch. Switching to another branch yields a 404.
    - In this case, fetching the new branch's tree would actually be pretty helpful.
- These changes are heavily GitHub-specific, and Bitbucket is completely disregarded. I opted into the Bitbucket "refreshed UI" and couldn't figure out how to opt out.
    - I'm happy to move as much logic into the GitHub adapter as possible, but at time of writing I'm turning in for the evening and want to get a round of feedback if possible.
- Should there be a branch filter as well? (Since I mentioned trying to match GitHub's functionality...)

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -52,6 +52,30 @@ class TreeView {
   _showHeader(repo) {
     const adapter = this.adapter
 
+    let currentBranchName = repo.branch.toString();
+    let branchHtml = '';
+
+    // Display branch switcher if the repository has multiple branches
+    if (repo.branches && repo.branches.length > 1) {
+      branchHtml = '<select id="octotree_header_branch_switcher">' +
+        repo.branches.map((branchName) => {
+          const safeBranchName = this._deXss(branchName).toString();
+
+          // Disable non-matching branch names, because URL won't work
+          const disabled = branchName !== safeBranchName;
+          const selected = branchName === currentBranchName;
+
+          return '<option value="' + safeBranchName + '"' +
+            (disabled ? ' disabled' : '') +
+            (selected ? ' selected' : '') +
+            '>' + safeBranchName + '</option>'
+        }).join('') +
+      '</select>';
+    }
+    else {
+      branchHtml = this._deXss(currentBranchName);
+    }
+
     this.$view.find('.octotree_view_header')
       .html(
         '<div class="octotree_header_repo">' +
@@ -60,7 +84,7 @@ class TreeView {
            '<a data-pjax href="/' + repo.username + '/' + repo.reponame + '">' + repo.reponame +'</a>' +
          '</div>' +
          '<div class="octotree_header_branch">' +
-           this._deXss(repo.branch.toString()) +
+           branchHtml +
          '</div>'
       )
       .on('click', 'a[data-pjax]', function (event) {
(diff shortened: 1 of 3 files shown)


buunguyen:
Sorry I thought I already provided feedback to this.

First, great work! Thanks for working on this.  A few things to consider:

* Improve the styling. Maybe match GitHub dropdown style?
* Avoid the 404 error as you mentioned. Maybe behave like GitHub branch selector itself?
* Would be great to avoid the disabled links. How does GitHub deal with that?

ChaseWagoner:
The "native" GitHub dropdown's behavior isn't specific to its hierarchy on the page, so we could just copy its HTML into this spot. That fixes all three issues, as it would then appear and behave exactly as the regular dropdown.

I don't imagine GitHub changes its branch switcher all too often, but this does create a pretty brittle relationship between the extension and GitHub's UI. What do you think?

ChaseWagoner:
After trying it out a bit, that's probably the answer. It looks great and behaves like a charm. Only thing I haven't worked out yet is the "branch" icon in `.octotree_header_branch:before`. It forces the branch switcher slightly out of line.

Without `:before` pseudoelement:

without

With `:before` pseudoelement:

with

The element really ought to be kept if we go this route, because plenty of pages don't have the branch switcher element to copy.

src/view.tree.js
```diff
@@ -52,28 +52,11 @@ class TreeView {
...
-            (disabled ? ' disabled' : '') +
-            (selected ? ' selected' : '') +
-            '>' + safeBranchName + '</option>'
-        }).join('') +
-      '</select>';
-    }
-    else {
-      branchHtml = this._deXss(currentBranchName);
+    // Copy GitHub's branch switcher
+    let branchHtml = $('<div>').append($('.branch-select-menu').clone()).html();
```
buunguyen: What is the page in question doesn't have the selector? Should we check if the element exist first and use fallback (i.e. no selector) if not? Also, should the element be styled/customized a bit so that it looks more fit in the octotree header?

igor-egorov:
Hi! Will that branch switcher recognize currently selected branch? 

Now clicking on any tree item will lead you to the file from the master branch, even when you were reviewing pull request or browsing repo on specific commit. :(

buunguyen:
I will need to rethink this feature and if want it, will need to have a better design.

PR #458, opened 2017-11-05
Added Width to Header Repo Text - Fixed Button Header UI

FlyteWizard:
### Description
Summary of what issue(s) this PR resolves

So this pull request solves a UI issue with the header repo and the button. The issue is mentioned here: https://github.com/buunguyen/octotree/issues/445

### Proposal
Summary of what changes this PR proposes

This pull request changes this: 
capture d ecran 2017-11-04 a 19 31 08

to this: 
capture d ecran 2017-11-04 a 19 31 36

I understand that this pull request is similar to https://github.com/buunguyen/octotree/pull/446, but they haven't submitted a new pull request in awhile. 

So the solution proposed in the above pull request was to limit the width of the header repo. I limited it to 152px.

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -73,6 +73,7 @@
 
     .octotree_treeview {
       .octotree_header_repo {
+        width: 152px;
         font-size: 13px;
         font-weight: normal;
       }


src/adapters/github.less
```diff
@@ -73,6 +73,7 @@
 
     .octotree_treeview {
       .octotree_header_repo {
+        width: 152px;
```
buunguyen: Not so great when ppl resize the sidebar, right?

FlyteWizard: Re-size the sidebar? How would someone do that?

FlyteWizard: Would you prefer

```css
width: 66%;
```

or do you have another option?

githubextension

buunguyen:
Ideally, the text should stay off the icons on the right, but not further (i.e. no cut off of text when people increase sidebar size).

PR #461, opened 2017-11-08
gulpfile.babel: fix the issue that only build on macOS

Ephemera:
### Description
I use `gulp-crx-pack` library and fake private key for chrome extension packaging.
If you have any ideas about private key or key location, tell me, please.


gulpfile.babel.js
```diff
@@ -80,15 +80,15 @@ gulp.task('chrome:zip', () => {
...
-  $.run('"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"' +
-    ' --pack-extension=' + path.join(__dirname, './tmp/chrome') +
-    ' --pack-extension-key=' + path.join(process.env.HOME, '.ssh/chrome.pem')
-  ).exec(cb)
-})
-
-gulp.task('chrome:crx', ['chrome:_crx'], () => {
-  return pipe('./tmp/chrome.crx', './dist')
+gulp.task('chrome:crx', () => {
+  const real = '.ssh/chrome.pem'
```
buunguyen: Should be `path.join(process.env.HOME, '.ssh/chrome.pem')`

test/fake.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
buunguyen: Suggest move this file to root directory instead of `test`, rename to `test_chrome_key.pem`

gulpfile.babel.js
buunguyen: This is better :) path.join(os.homedir(), '.ssh/chrome.pem')

buunguyen:
Sorry for the late response, been traveling. Just some minor suggestions.

Ephemera:
@buunguyen 

It's okay. Was your travel good? 
I fixed your suggestions.

buunguyen:
Thanks for taking long. Just merged!

PR #477, opened 2018-01-15
Fix file icons being in italic

Akryum:
### Description
Currently, the file icons are in italic.

### Proposal
This add the `font-style: normal` to the icons.


buunguyen:
Thanks!

Issue #480, opened 2018-01-16
Expecting a port to Microsoft Edge


lizy14:

buunguyen:
Seems reasonable as Edge extensions are Chrome-compatible. I don't develop on Windows and wouldn't have time for this anyway. Anyone wants to take on?

Jack-Q:
Hi @buunguyen, I have made some attempts to port is to Edge. 
Currently, it basically works well in Edge, while with the following features are broken:

* file-specific icons
* expanding animation of js-tree

Here is a preview of a GitHub page with Octotree in Edge:
![image](https://user-images.githubusercontent.com/9014072/35631359-ab4557de-06de-11e8-8f4a-a2484cc5997b.png)

Jack-Q:
Update:
Currently, the animation of js-tree works fine in the latest version of Edge. And the file-specific icon is still blocked due to the failure of icon font loading (Caused by an issue of Edge [https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11320212/]()).

NitroAMD:
@Jack-Q - Thank you for your efforts here. The latest Windows 10 1803 update should fixed the Edge issues. I appreciate your work here and hope to see Octotree for Edge soon. :-)

lizy14:
May I ask the progress please? When can I expect a binary release to sideload, and further, listing in Microsoft Store?

buunguyen:
There's a PR here: https://github.com/buunguyen/octotree/pull/488.
I'm busy right now and I can't install VM/Edge and test it out. If anyone can test the Edge PR locally and confirm it works then I can merge it.

NitroAMD:
I tried to use PR #488 but it don't work for me at all. Any thoughts?
https://docs.microsoft.com/en-us/microsoft-edge/extensions/guides/adding-and-removing-extensions#adding-an-extension

![image](https://user-images.githubusercontent.com/1763966/40922551-c310ef6e-6812-11e8-8789-1f587f6ff7be.png)

Jack-Q:
Hi @NitroAMD, this project needs to be built before sideload. You can use `gulp edge` command to build it against Edge (don't forget `npm install`). Then, you can sideload the folder `tmp\edge\Extension`. 

After building, the version number in the manifest will be replaced by the real one and shown in the extension info page. Here is a screenshot showing that.
[...]

PR #488, opened 2018-01-31
Port to Microsoft Edge 

Jack-Q:
### Description

A port of this extension to Microsoft Edge.

This is port contains some modifications to the original extension to some compatibility issues in Microsoft Edge.  #480


src/octotree.js
```diff
@@ -190,4 +190,10 @@ $(document).ready(() => {
       return $toggler.is(':visible')
     }
   }
-})
+}
+
+if(document.readyState === 'complete') {
```
buunguyen: `$(document).ready(...)` not work? I'm quite surprised if that's the case.

nitpick: need a space after `if`.

src/config/edge/manifest.json
```diff
@@ -0,0 +1,41 @@
+{
+    "name": "Octotree",
+    "version": "$VERSION",
+    "manifest_version": 2,
+    "author": "Buu Nguyen",
+    "description": "Code tree for GitHub",
+    "icons": {
```
buunguyen: Which manifest file Edge uses? Why icons are defined in both files?

src/view.tree.js
```diff
@@ -1,5 +1,6 @@
 class TreeView {
   constructor($dom, store, adapter) {
+    const isEdge = window.navigator.userAgent.indexOf('Edge') > 0;
```
buunguyen: Would be great if there's a way to make it work without specific browser detection & handling like this.

gulpfile.babel.js
```diff
@@ -148,6 +148,27 @@ gulp.task('safari', ['safari:js'], () => {
   )
 })
 
+// Edge
+gulp.task('edge:template', () => {
+  return buildTemplate({SUPPORT_FILE_ICONS: true})
```
buunguyen: You mentioned in the issue file icons aren't working. If that's the case, should mark this as 'false' until there's a fix. That being said, I'd love to have that fixed before merging, unless Edge can't really support fonts.

iRYO400:
Thumb up! Waiting for Edge version!

src/config/edge/manifest.json
Jack-Q: Actually, Edge extension is packaged as `appx` and published to Microsoft Store. Both manifests are used.

* `manifest.json` is used by Edge and the icon is displayed within Edge;
* `AppxManifest.xml` is used by Microsoft Store and the icon is displayed in the Marketplace.

They should be the same icon but size requirement of them are not the same.

src/octotree.js
Jack-Q: I don't know the reason why `$(document).ready(...)` fails either.

src/view.tree.js
Jack-Q: It seems that this problem is resolved in the latest version of Edge. I have removed the browser detection code.

gulpfile.babel.js
Jack-Q: The support configuration is removed for Edge. Actually, this is caused by an issue impacting how Edge handles Content Security Policy headers, which blocks the loading of icon fonts. This issue is fixed but not flighted yet. (Related Issue: [https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11320212/]())

jamesrweb:
@buunguyen @Jack-Q any movement upcoming to this, would be good to get things merged if all minor topics (seem to be) resolved?

piranna:
Microsoft Edge has got deprecated, I think this can be rejected.

abulka:
Now that Microsoft Edge has been reborn with a chrome engine - is there any reason why Octotree isn't available as an extension for Microsoft Edge?

buunguyen:
Edge version: https://microsoftedge.microsoft.com/addons/detail/octotree/joagmknfcgpikbadjkaikmnhpjadihjg?hl=en-US

PR #492, opened 2018-02-08
Easier token generation on Github

Jackymancs4:
### Description
Inspired by other github-related extensions (notably OctoLinker browser extension and OctoLinker) this PR aims to automate even more token generation providing the minimal preset of permissions.

**Before:**
https://github.com/settings/tokens/new

**After:**
https://github.com/settings/tokens/new?scopes=public_repo&description=OctoTree%20browser%20extension

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -58,7 +58,7 @@ class GitHub extends PjaxAdapter {
 
   // @override
   getCreateTokenUrl() {
-    return `${location.protocol}//${location.host}/settings/tokens/new?scopes=public_repo&description=OctoTree browser extension`
+    return `${location.protocol}//${location.host}/settings/tokens/new?scopes=repo&description=Octotree%20browser%20extension`
   }
 
   // @override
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -58,7 +58,7 @@ class GitHub extends PjaxAdapter {
 
   // @override
   getCreateTokenUrl() {
-    return `${location.protocol}//${location.host}/settings/tokens/new`
+    return `${location.protocol}//${location.host}/settings/tokens/new?scopes=public_repo&description=OctoTree browser extension`
```
buunguyen: Good idea. I think many people only start realizing they need token when they hit a private repo and Octotree prompts for the token. Therefore, maybe a good idea to automatically select all repos to begin with. Also, please change description to "Octotree browser extension".

Jackymancs4: Good points, done!

buunguyen:
Thanks!

PR #496, opened 2018-03-08
Support file icons in Safari

### LINKED ISSUE #493: File icons always same between file and directory [Safari]
### Description
File icons always same between file and directory in **Safari**

### Environment (if bug)

* Octotree version: 2.4.3
* Browser & version: Safari 11.0.3
* OS & version: macOS 10.13.3
* Screenshot, if any (drag an image here)

screen shot 2018-02-12 at 2 16 52 pm

tbodt:
Fixes #493

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -129,11 +129,11 @@ gulp.task('firefox:zip', () => {
 
 // Safari
 gulp.task('safari:template', () => {
-  return buildTemplate({})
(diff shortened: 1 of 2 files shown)


buunguyen:
Have you tested it? Screenshot would be great.

buunguyen:
You're right. I also thought it needed special handling for the font path like the case with Chrome and Firefox. Turns out it's quite simple. Thanks for working on this!

PR #501, opened 2018-03-22
fix(explorer): restore focus after href scroll

cogell:
### Description
when using the keyboard to navigate files inside octotree, inside a PR,
the octotree explorer would lose focus on scroll

### Proposal
this code fixes that issue restoring focus after document scroll

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -120,10 +120,12 @@ class TreeView {
     if (!$target.is('a.jstree-anchor')) return
 
     // refocus after complete so that keyboard navigation works, fix #158
+    // `scroll` event fires when selecting files in PR which scrolls to new #href in page
     const refocusAfterCompletion = () => {
-      $(document).one('pjax:success page:load', () => {
-        this.$jstree.get_container().focus()
-      })
+      $(document)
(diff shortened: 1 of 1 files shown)


src/view.tree.js
```diff
@@ -121,9 +121,13 @@ class TreeView {
 
     // refocus after complete so that keyboard navigation works, fix #158
     const refocusAfterCompletion = () => {
-      $(document).one('pjax:success page:load', () => {
-        this.$jstree.get_container().focus()
-      })
+      const refocus = () => this.$jstree.get_container().focus();
+      $(document)
+        .one('pjax:success page:load', () => { // happens when navigating between files
```
akread: Is there a reason you can't just add the scroll event to the event string code that already exists? Multiple events are separated by spaces. 🤔 

```js
$(document).one('pjax:success page:load scroll', () => {
```

cogell: oh, good point!

cogell:
heads up: if you click a new file it will restore focus and then restore focus again when you scroll in that new page. dunno if this is acceptable...

buunguyen:
Sorry, I'm not clear about the benefit vs the changed behavior, so gonna close.

PR #502, opened 2018-03-22
WIP: feat(toggle-show): toggle focus on show

cogell:
### Description
when toggling hide/show of the octotree explorer the users cursor focus doesnt change

### Proposal
when toggling hide/show of the octotree explorer the users cursor focuses into the tree on show and blurs on hide


cogell:
issue: on initial load the tree has a focus NOT correctly in the tree

buunguyen:
Why should it blur on hide?

PR #509, opened 2018-04-23
Ignore case when sorting

Ju2ender:
### Description
Ignore case when sorting. Help us find the file quickly.

Before:
octotree1

After:
octotree2


buunguyen:
The current sort matches GitHub own tree sort order. I'd prefer to keep it that way.

Ju2ender:
Alright.

PR #512, opened 2018-04-28
Get current user info from meta tag instead of body

### LINKED ISSUE #508: VM559:1 Uncaught SyntaxError: Unexpected token u in JSON at position 0
### Description
It appears it's not getting the username from bitbucket.  Rather, body does not have an attribute data-current-user.  Could a change in site design have caused this or does data-current-user get rendered within octotree? 

### Environment (if bug)

* Octotree version: 2.4.6
* Browser & version: 65.0.3325.181 (Official Build) (64-bit)
* OS & version: macOS High Sierra 10.13.4
* Screenshot:
screen shot 2018-04-18 at 1 26 09 pm

jamez14:
### Description
Fixes #508. Bitbucket changed how where they render the `data-current-user` attribute. It has been moved from the `body` tag to its own `meta` tag directly inside the `<head>`. 

Here is what it looks like now:

screen shot 2018-04-28 at 9 47 28 am

### Proposal
Instead of selected the `data-current-user` from the `body`, grab it from the `meta` instead.

### DIFF
--- a/src/adapters/bitbucket.js
+++ b/src/adapters/bitbucket.js
@@ -180,7 +180,7 @@ class Bitbucket extends PjaxAdapter {
         username =  result[0], token = result[1]
       }
       else {
-        const currentUser = JSON.parse($('body').attr('data-current-user'))
+        const currentUser = JSON.parse($('meta').attr('data-current-user'))
         if (!currentUser || !currentUser.username) {
           return cb({
             error: 'Error: Invalid token',


buunguyen:
Thank you!

dmbaranov:
True, would be good if author can update extension in the Chrome Web Store. Until that, you can download the source code and install unpacked extension yourself. Just run `yarn build`, after that `yarn dist`. Once it's finished, go to the dist folder and extract chrome.zip. When done, go to chrome://extensions, turn on Developer mode, press Load unpacked button and select the extracted chrome folder. Should work like a charm :)

P.S. I assume that you're @jamez14 an experienced developer and this guide might be no that helpful for you. But it could be helpful for someone who'll find this pull request :)

PR #518, opened 2018-05-31
Dark mode added

omidnikrah:
Dark mode toggle added to settings 🚀


dantehemerson:
Looks good :+1:

anguianomaria:
Hola  yo ni entiendo esto de que se trata

ttuygun:
@buunguyen can we merge this request?

buunguyen:
Thanks for this PR. Unfortunately, I cannot merge because:

* The code in this PR isn't consistent with existing code. This could be updated but the bigger thing is: 
* I'm thinking of building theming support of Octotree. And it would also control GitHub itself, not just the tree.

Issue #339, opened 2016-10-24
Looks like Download button is present but it isn't working


karuppiah7890:
@buunguyen 
I can see a download icon when I hove over the left side of the file names. But I ain't able to download them when I click the download icon.

![error](https://cloud.githubusercontent.com/assets/12808424/19654725/347ddc8a-9a37-11e6-9146-f46bfd0dde50.png)

buunguyen:
What browser/OS is that?

karuppiah7890:
Mozilla Firefox v51.0a2 in Ubuntu 16.10

buunguyen:
Yeah, looks like a bug, happens in OSX/Firefox too. PR is welcome.

scilganon:
Confirm: don't work
OS: Win10
Ver: 49.0.2

----

found solutions:

- https://support.mozilla.org/en-US/questions/968992 - just append created link into document
- http://stackoverflow.com/a/27116581
- http://stackoverflow.com/a/30838895 - convert to blob

IvanJov:
@scilganon Hey, were you able to fix this issue? I am able to reproduce it and I can help you with it. Let me know :)

buunguyen:
@IvanJov check this out https://github.com/buunguyen/octotree/pull/352

shaojiankui:
same issue macos firefox 50.1.0

XCanG:
I have similar issue on Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0 ID:20180124100321
If you use <a> links probably you may need download parameter: `<a download="README.md"></a>` to download it on click, or in JS it equivalent to:
```javascript
a.setAttribute('download', 'README.md');
a.setAttribute('href', 'https://raw.githubusercontent.com/buunguyen/octotree/master/README.md');
a.click();
```

PR #547, opened 2018-10-10
Fix bug file downloading

duylam:
Fix https://github.com/ovity/octotree/issues/339

Tested:
- Chrome on Mac OS
- Firefox on Mac OS


buunguyen:
Can you replace the download icon too? There are many icon fonts included, so there should be one that better communicates the intent of "open file in new tab".

duylam:
@buunguyen Addressed the comment, the new icon is https://octicons.github.com/icon/browser/, below is the screenshot

screen shot 2018-10-12 at 22 31 36

duylam:
@buunguyen Addressed comments

Issue #514, opened 2018-05-04
Option to disable on Bitbucket


djekl:
### Description
Would like the option to disable on BitBucket as I found it too intrusive, but perfect for GitHub

crashbell:
There are some issues I saw on Bitbucket, I would like to fix them. Can you please list down issues you saw on Bitbucket? So I can fix them too. Disabling it in a specific website could be an option, but I think we need to fix the issues instead of ignoring them by disable option.

djekl:
The main issue is its placement, where it is, you can't press the bitbucket icon to go back to home. I don't like it on BitBucket, so an option to disable it on the site is perfect for me. The option to disable it on a specific site (for example bitbucket, or a corporate GitHub instance) isn't a band aid, its more to have the option to disable it on there.

kylef000:
@crashbell Similarly, I love OctoTree on Github but I'd rather not have to see it on my work BitBucket.

![image](https://user-images.githubusercontent.com/1134391/41164271-bd4bca2e-6b08-11e8-9a3f-769635da6391.png)

efemoney:
Bumping this. I have the same issue.

PR #407, opened 2017-05-13
Support bitbucket [Building upon #328]

tvytlx:
### Description
Fix all requests for improvements in #328.

### Proposal
Support bitbucket

PR #550, opened 2018-10-13
Remove supporting Bitbucket

duylam:
### Problem

Address https://github.com/ovity/octotree/issues/514 by removing Bitbucket

### Solution

- Revert changes for Bitbucket at PR #407
- Tested on Chrome and Firefox at repos at bitbucket.org


aitbahaa:
Why no longer supporting bitbucket?

PR #552, opened 2018-10-15
Update GitHub repo parser to handle browsing repo at commit. Fixes #473.

### LINKED ISSUE #473: Links go to master branch from commit tree page in github
### Description
In github.com when select "commit", in sidebar links go to master branch but not to this commit files.
For example go to commit https://github.com/buunguyen/octotree/commit/b06e421795b403d46939055c30a5c1e1b111c706 files tree
https://github.com/buunguyen/octotree/tree/b06e421795b403d46939055c30a5c1e1b111c706

### Environment (if bug)

* Octotree version: 2.3.3
* Browser & version: Google Chrome 63.0.3239.132 x64
* OS & version: Mac OS Hight Sierra 10.13.2 (17C88)

stevennoto:
### Problem
This addresses #473, where Octotree was not handling viewing the repo at commit level properly. In that case, it was not finding any branch or commit info, and falling back to `master`, and showing repo contents and links based on `master` instead of at the current commit.

### Solution
This solution makes use of the fact that GitHub URL structure and API endpoints treat branches, labels, and sha commit references equivalently. The code now recognizes the commit sha from the URL and uses it as the branch. Therefore the sha is displayed in Octotree, used for the repo API call, and is present in all links generated.

### Screenshots
This shows Octotree showing a repo at a commit:
screen shot 2018-10-15 at 12 09 50 am

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -118,14 +118,14 @@ class GitHub extends PjaxAdapter {
     const pullNumber = isPR && showOnlyChangedInPR ? typeId : null
 
     // Skip non-code page unless showInNonCodePage is true
-    if (!showInNonCodePage && type && !~['tree', 'blob', 'commit'].indexOf(type)) {
+    const isCodePage = ['tree', 'blob', 'commit'].indexOf(type) >= 0
+    if (!showInNonCodePage && type && !isCodePage) {
       return cb()
     }
 
-    const isBranchOrSha = ['tree', 'blob', 'commit'].indexOf(type) >= 0 && typeId
-    const branchOrSha = isBranchOrSha && typeId ? typeId : null
-
     // Get branch by inspecting page, quite fragile so provide multiple fallbacks
+    const isBranchOrSha = isCodePage && typeId
+    const branchOrSha = isBranchOrSha && typeId ? typeId : null
     const branch =
       // Branch/Tag/Commit from URL
       branchOrSha ||
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -118,12 +118,17 @@ class GitHub extends PjaxAdapter {
     const pullNumber = isPR && showOnlyChangedInPR ? typeId : null
 
     // Skip non-code page unless showInNonCodePage is true
-    if (!showInNonCodePage && type && !~['tree', 'blob'].indexOf(type)) {
+    if (!showInNonCodePage && type && !~['tree', 'blob', 'commit'].indexOf(type)) {
       return cb()
     }
 
+    const isBranchOrSha = ['tree', 'blob', 'commit'].indexOf(type) >= 0 && typeId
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -118,12 +118,17 @@ class GitHub extends PjaxAdapter {
     const pullNumber = isPR && showOnlyChangedInPR ? typeId : null
 
     // Skip non-code page unless showInNonCodePage is true
-    if (!showInNonCodePage && type && !~['tree', 'blob'].indexOf(type)) {
+    if (!showInNonCodePage && type && !~['tree', 'blob', 'commit'].indexOf(type)) {
```
duylam: Hi @stevennoto, the PR looks nice, just one minor improvement. How about introduce a variable for `['tree', 'blob', 'commit'].indexOf(type) >= 0` to represent for "current page is code page" ?

stevennoto: Great idea! I'll fix this tonight.

stevennoto:
@duylam, I added the variable you mentioned, and tried to group the new lines better with the existing lines. See what you think. Thanks!

src/adapters/github.js
```diff
@@ -118,12 +118,17 @@ class GitHub extends PjaxAdapter {
     const pullNumber = isPR && showOnlyChangedInPR ? typeId : null
 
     // Skip non-code page unless showInNonCodePage is true
-    if (!showInNonCodePage && type && !~['tree', 'blob'].indexOf(type)) {
+    const isCodePage = ['tree', 'blob', 'commit'].indexOf(type) >= 0
+    if (!showInNonCodePage && type && !isCodePage) {
```
buunguyen: If type is not undefined, then it's code page. So this should be:
```
const isCodePage = !type || ['tree', 'blob', 'commit'].indexOf(type) >= 0
if (!showInNonCodePage && !isCodePage)
```

src/adapters/github.js
```diff
@@ -118,12 +118,17 @@ class GitHub extends PjaxAdapter {
...
       return cb()
     }
 
     // Get branch by inspecting page, quite fragile so provide multiple fallbacks
+    const isBranchOrSha = isCodePage && typeId
+    const branchOrSha = isBranchOrSha && typeId ? typeId : null
```
buunguyen: The second `typeId` check isn't necessary. Should just merge these 2 lines to: 
```
const branchOrSha = isCodePage && typeId
```
But really, that's the whole point of the "branch" variable. Might just combine:

```
const branch = 
   // Branch/tag/commit from URL
   (isCodePage && typeId) ||
   // Code page DOM
   // .. the rest
```

stevennoto:
Thanks for the feedback @buunguyen! I made the changes you requested.

buunguyen:
Looking good. Thanks for working on this @stevennoto!

PR #556, opened 2018-10-21
Forward view's events to plugins

duylam:


buunguyen:
Why? Plugins can listen to events directly with $(aView).on(EVENT.VIEW_CLOSE). No need to forward like this.

Issue #559, opened 2018-10-23
"Error: Private Repository" on non-master branches


coraWhite:
### Description
Octotree sidebar reports "Error: Private Repository" prompting to create a new access token on any branch for a Private Repository besides Master branch. Works normally on a Master branch. 

This started the day I updated Octotree to 2.5.0 and before Firefox was updated to 63.0. After updating Firefox to 63.0, same error.

I deleted previous access tokens and re-created a new one with no luck.

### Environment (if bug)

* Octotree version:  2.5.0
* Browser & version: Firefox 63.0 
* OS & version: Mac OSX 10.13.6

leonard-waters:
Same error occurring on Chrome as well.

- Octotree Version: 2.5.0
- Browser & Version: Chrome 69.0.3497.100
- OS & Version: Mac OSX 10.14.1

duylam:
@leonard-waters

> Took a quick look at this and it looks like the issue involves the request to Get a tree recursively. Only the first part of the branch name is being added to the query.

Thanks for the hint, it's correct that the mishandling branch name is the problem. The fixing code would be at https://github.com/ovity/octotree/blob/master/src/adapters/github.js#L124, I will explore more cases on url and apply a fix

buunguyen:
Fixed in 2.5.1.

leonard-waters:
Similar error again, but not exactly the same. Naming convention does not seem to be related in this case.

coraWhite:
This issue has reoccurred for me as well, with versions:
Octotree: 2.5.1
Firefox: 64.0.2 (64-bit)
MacOS Mojave: 10.14.2

And the naming convention seems to be a consistent issue for me particularly. A non-master branch, such as `user` on a private repo will load correctly. However, a non-master branch containing `-`, `/`, or `_` does not load correctly on the same private repo.

philmayfield:
Octotree: 2.5.1
Chrome: 71.0.3578.98
MacOS Mojave: 10.14.2

I'm getting a similar error around naming of branches.  Branches with only characters appear to work fine with Octotree.  When using the `-` and `_` characters is where I'm seeing the error.  I tried for a bit to come up with a rhyme or reason, but best I can figure is its some combination of length and those special characters that is throwing it.  Very short branch names with them work ok, but longer names do not.

llrs:
I confirm that after renaming the branch (without "_" on my case) the issue is solved and the files are shown

andersrehn:
[...]

PR #564, opened 2018-10-26
Fix issue #559

duylam:
Fix https://github.com/ovity/octotree/issues/559

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -120,16 +120,17 @@ class GitHub extends PjaxAdapter {
       return cb();
     }
 
-    // (username)/(reponame)[/(type)][/(typeId)]
-    const match = window.location.pathname.match(/([^\/]+)\/([^\/]+)(?:\/([^\/]+))?(?:\/([^\/]+))?/);
-    if (!match) {
+    // (username)/(reponame)[/(type)][/(typeId1)][/(typeId2)][...]
+    const match = window.location.pathname.split('/').filter((part) => (!!part));
+
+    if (!match.length) {
       return cb();
     }
 
-    let username = match[1];
-    let reponame = match[2];
-    let type = match[3];
-    let typeId = match[4];
+    let username = match[0];
+    let reponame = match[1];
+    let type = match[2];
+    let typeId = match[3];
 
     // Not a repository, skip
     if (~GH_RESERVED_USER_NAMES.indexOf(username) || ~GH_RESERVED_REPO_NAMES.indexOf(reponame)) {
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -161,7 +162,57 @@ class GitHub extends PjaxAdapter {
...
+        });
+
+        this._get('/tags', {repo, token}, (err, data) => {
+          repoTags = err ? [] : data.map((item) => (item.name));
+
+          onResponseAvailable(err);
+        });
+
+        function onResponseAvailable(err) {
+          if (!repoBranches || !repoTags) return;
```
buunguyen: Should be done with util#parallel.

src/adapters/github.js
```diff
@@ -161,7 +162,57 @@ class GitHub extends PjaxAdapter {
...
+        //  - /username/repo/tree (or blob)/branch1/folder1
+        //  - /username/repo/tree (or blob)/features/a (Gitflow workflow)
+        //  - /username/repo/tree (or blob)/features/a/folder1
+        // Reserve the branch name which is possibly mixing with the file path
+        const typeIdWithPath = match.slice(3);
+        let repoBranches, repoTags;
+
+        this._get('/branches', {repo, token}, (err, data) => {
+          repoBranches = err ? [] : data.map((item) => (item.name));
+
```
buunguyen: Why are all these blank lines?

src/adapters/github.js
```diff
@@ -161,7 +162,57 @@ class GitHub extends PjaxAdapter {
...
+          if (err) return cb(err);
+
+          const gitSnapshotNames = repoBranches.concat(repoTags);
+
+          let branchNameToTry = '';
+          let foundBranch = false;
+
+          // Verify the branch / tag name by comparing with all branches and tags from the repo
+          typeIdWithPath.forEach((typeIdPart, index) => {
```
buunguyen: I don't get this. Isn't the first element of `typeIdWithPath` the branch/tag already? What is this loop about?

src/adapters/github.js
buunguyen: Very concerned about these 2 calls. They happen in normal flow instead of exception flow. So network is hit every single time. Costly IMO. 

1. Isn't there any way to do what you're doing without hitting the network?
2. If has to hit network, must structure the code so that this is the last resort. For example, if `$('.branch-select-menu .select-menu-item.selected').data('name')` returns value, we have the branch name. Why have to go through all these extra network requests.

PR #565, opened 2018-10-27
Fix: get 404 when branch name has slash

nphi1212:
Suggest another solution for the issue [#559](https://github.com/ovity/octotree/issues/559)

a @duylam , Could you check if I might miss any case ?

Many thanks.

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -146,12 +146,15 @@ class GitHub extends PjaxAdapter {
       return cb();
     }
 
+    const branchOnMenu = $('.branch-select-menu .select-menu-item.selected').data('name') || '';
+
     // Get branch by inspecting URL or DOM, quite fragile so provide multiple fallbacks
     const branch =
       // Branch/tag/commit from URL
-      (isCodePage && typeId) ||
+      // Exclude edge case: branch name has slash '/'
+      (!branchOnMenu.includes('/') && isCodePage && typeId) ||
       // Code page
-      $('.branch-select-menu .select-menu-item.selected').data('name') ||
+      branchOnMenu ||
       // Pull requests page
       ($('.commit-ref.base-ref').attr('title') || ':').match(/:(.*)/)[1] ||
       // Reuse last selected branch if exist
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -146,15 +146,12 @@ class GitHub extends PjaxAdapter {
       return cb();
     }
 
-    const branchOnMenu = $('.branch-select-menu .select-menu-item.selected').data('name') || '';
-
     // Get branch by inspecting URL or DOM, quite fragile so provide multiple fallbacks
     const branch =
-      // Branch/tag/commit from URL
-      // Exclude edge case: branch name has slash '/'
-      (!branchOnMenu.includes('/') && isCodePage && typeId) ||
+      // Code page with commit ID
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -146,12 +146,15 @@ class GitHub extends PjaxAdapter {
...
     }
 
+    const branchOnMenu = $('.branch-select-menu .select-menu-item.selected').data('name') || '';
+
     // Get branch by inspecting URL or DOM, quite fragile so provide multiple fallbacks
     const branch =
       // Branch/tag/commit from URL
-      (isCodePage && typeId) ||
+      // Exclude edge case: branch name has slash '/'
+      (!branchOnMenu.includes('/') && isCodePage && typeId) ||
```
buunguyen: Or maybe we should use `typeId` only when type is 'commit'?

buunguyen: How about we only use `typeId` when it's a url is `commit`? For others (`tree` and `blob`), just use `branchOnMenu`? What is simpler to understand and more future proof?

The reason I don't quite love the fix in this PR is it combines both URL and DOM to make a determination. 2 volatile things are more volatile in combination :).

src/adapters/github.js
```diff
@@ -146,10 +146,13 @@ class GitHub extends PjaxAdapter {
       return cb();
     }
 
+    // Trivial check whether typeId is a commit id in case URL '.../tree/{commitId}'
```
buunguyen: in case URL '.../commit/{commitId}'

src/adapters/github.js
```diff
@@ -146,10 +146,13 @@ class GitHub extends PjaxAdapter {
...
+    const isCommit = type === 'commit' || !!typeId.match(/[a-z, 0-9]{40}/)
```
buunguyen: Not sure if we need the 2nd check. I think it could just be

```
const branch = 
   // Special handling for commit
   type === 'commit' && typeId ||
   // Code page
   ...

File: src/adapters/github.js (hunk 2 above)
Comment (nphi1212) [reply src/adapters/github.js]: I mean this case https://github.com/ovity/octotree/tree/b06e421795b403d46939055c30a5c1e1b111c706

File: src/adapters/github.js [hunk 4]
```diff
@@ -148,8 +148,8 @@ class GitHub extends PjaxAdapter {
 
     // Get branch by inspecting URL or DOM, quite fragile so provide multiple fallbacks
     const branch =
-      // Branch/tag/commit from URL
```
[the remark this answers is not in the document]
buunguyen: sgtm, the isCodePage check isn't necessary if type==='commit' tho.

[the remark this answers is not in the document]
nphi1212: @duylam. Thanks for your suggestion. I have tried that approach but it won't cover this case when browsing the commit in the tree page.  https://github.com/ovity/octotree/tree/14a07088a517de16b1c9e78b5b50bf42d45f3c37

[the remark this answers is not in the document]
duylam: @nphi1212 cc. @buunguyen  Please check https://github.com/nphi1212/octotree/pull/1

nphi1212:
I have merged a @duylam  enhancement. I think this PR is good to go. a @buunguyen

buunguyen:
LGTM. Thanks @nphi1212!

PR #568, opened 2018-11-03
Feature: Sidebar v2

nphi1212:
Changes:
- Unpinned mode:   Open/close sidebar by hover ( Sidebar overlay Github page )
screen shot 2018-11-03 at 11 15 44 pm
- Pinned mode: always open ( Sidebar push Github page )
screen shot 2018-11-03 at 11 15 50 pm
- Shortcuts: replace Shown sidebar shortcut with Pin shortcut
- Remove remember sidebar visibility option

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ By default, Octotree only works on `github.com`. To support enterprise version (
 
 ### Others
 * __Hotkeys__: Octotree uses [keymaster](https://github.com/madrobby/keymaster) to register hotkeys. Check out the [supported keys](https://github.com/madrobby/keymaster#supported-keys).
-* __Remember sidebar visibility__: if checked, show or hide Octotree based on its last visibility.
 * __Show in non-code pages__: if checked, allow Octotree to show in non-code pages such as Issues and Pull Requests.
 * __Load entire tree at once__: if checked, load and render the entire code tree at once. To avoid long loading, this should be unchecked if you frequently work with very large repos.
 * __Show only pull request changes__ _(new!)_: if checked and in "Pull requests" page, only show the change set of the pull request.
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -86,23 +86,17 @@ class GitHub extends PjaxAdapter {
   }
 
   // @override
-  updateLayout(togglerVisible, sidebarVisible, sidebarWidth) {
+  updateLayout(sidebarPinned, sidebarVisible, sidebarWidth) {
     const SPACING = 10;
     const $header = $(GH_HEADER);
     const $containers = $(GH_CONTAINERS);
     const autoMarginLeft = ($(document).width() - $containers.width()) / 2;
-    const shouldPushEverything = sidebarVisible && autoMarginLeft <= sidebarWidth + SPACING;
+    const shouldPushEverything = sidebarPinned && sidebarVisible;
+    const smallScreen = autoMarginLeft <= sidebarWidth + SPACING;
 
-    $('html').css('margin-left', shouldPushEverything ? sidebarWidth : '');
-    $containers.css('margin-left', shouldPushEverything ? SPACING : '');
-
-    const headerPadding =
-      shouldPushEverything || (!togglerVisible && !sidebarVisible)
-        ? '' // Nothing is visible or already pushed, leave as-is
-        : !sidebarVisible
-          ? 25 // Sidebar is collapsed, move the logo to avoid hiding the toggler
-          : sidebarWidth; // Otherwise, move the header from the sidebar
-    $header.css('padding-left', headerPadding);
+    $('html').css('margin-left', shouldPushEverything && smallScreen ? sidebarWidth : '');
+    $containers.css('margin-left', shouldPushEverything && smallScreen ? SPACING : '');
+    $header.css('padding-left', shouldPushEverything && !smallScreen ? sidebarWidth : '');
   }
 
   // @override
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -5,13 +5,38 @@
 
   .octotree_github_sidebar {
     a.octotree_toggle {
-      right: 12px;
-      top: 14px;
+      visibility: hidden;
+    }
 
-      &:not(.octotree_loading) > span:after,
-      &:not(.octotree_loading):hover > span:after {
-        #octicons.chevron-left();
-        font-size: 15px;
+    a.octotree_spin {
+      right: 12px;
+      top: 65px;
+      &.octotree_loading {
+        & > span:after {
+          content: '';
+        }
+
+        .loader {
+          border-radius: 50%;
+          border: 2px solid #000;
+          border-top: 2px solid rgba(0, 0, 0, 0);
+          border-left: 2px solid rgba(0, 0, 0, 0);
+          width: 16px;
+          height: 16px;
+          animation: loading 0.5s infinite linear;
+        }
+      }
+        @keyframes loading {
+        0% {
+          transform: rotate(0deg);
+        }
+        100% {
+          transform: rotate(360deg);
+        }
+      }
+
+      &:not(.octotree_loading) {
+        visibility: hidden;
       }
     }
   }
@@ -159,7 +184,7 @@
 
   a.octotree_opts {
     top: 21px;
-    right: 48px;
+    right: 12px;
     width: 14px;
     height: 16px;
 
@@ -174,19 +199,41 @@
     }
   }
 
+  a.octotree_pin {
+    top: 21px;
+    right: 36px;
+    width: 14px;
+    height: 16px;
+
+    &:hover i:before {
+      color: @blue;
+    }
+
+    & i:before {
+      #octicons.pin();
+      color: @dark;
+      font-size: 15px;
+    }
+  }
+
+  a.octotree_pinned i:before {
+    #octicons.pin();
+    transform: rotate(-45deg);
+  }
+
   a.octotree_toggle {
-    top: 14px;
+    top: 50%;
     right: -35px;
 
-    &:not(.octotree_loading) > span:after,
-    &:not(.octotree_loading):hover > span:after {
+    & > span:after,
+    &:hover > span:after {
       #octicons.chevron-right();
       font-size: 15px;
       position: relative;
       left: 3px;
     }
 
-    &:not(.octotree_loading):hover > span:after {
+    &:hover > span:after {
       color: @blue;
     }
   }
--- a/src/core.constants.js
+++ b/src/core.constants.js
@@ -1,10 +1,11 @@
 const NODE_PREFIX = 'octotree';
 const ADDON_CLASS = 'octotree';
 const SHOW_CLASS = 'octotree-show';
+const PINNED_CLASS = 'octotree_pinned';
+const SIDEBAR_HIDING_DELAY = 3000;
 
 const STORE = {
   TOKEN: 'octotree.access_token',
-  REMEMBER: 'octotree.remember',
   NONCODE: 'octotree.noncode_shown',
   PR: 'octotree.pr_shown',
   HOTKEYS: 'octotree.hotkeys',
@@ -12,13 +13,12 @@ const STORE = {
   LOADALL: 'octotree.loadall',
   POPUP: 'octotree.popup_shown',
   WIDTH: 'octotree.sidebar_width',
-  SHOWN: 'octotree.sidebar_shown',
-  GHEURLS: 'octotree.gheurls.shared'
+  GHEURLS: 'octotree.gheurls.shared',
+  PINNED: 'octotree.sidebar_pinned'
 };
 
 const DEFAULTS = {
   TOKEN: '',
-  REMEMBER: true,
   NONCODE: true,
   PR: true,
   LOADALL: true,
@@ -32,6 +32,7 @@ const DEFAULTS = {
 
 const EVENT = {
   TOGGLE: 'octotree:toggle',
+  PIN: 'octotree:pin',
   LOC_CHANGE: 'octotree:location',
   LAYOUT_CHANGE: 'octotree:layout',
   REQ_START: 'octotree:start',
@@ -39,5 +40,6 @@ const EVENT = {
   OPTS_CHANGE: 'octotree:change',
   VIEW_READY: 'octotree:ready',
   VIEW_CLOSE: 'octotree:close',
-  FETCH_ERROR: 'octotree:error'
+  FETCH_ERROR: 'octotree:error',
+  HOTKEYS_CHANGED: 'octotree:change'
 };
--- a/src/octotree.js
+++ b/src/octotree.js
@@ -9,7 +9,9 @@ $(document).ready(() => {
     const $dom = $(TEMPLATE);
     const $sidebar = $dom.find('.octotree_sidebar');
     const $toggler = $sidebar.find('.octotree_toggle');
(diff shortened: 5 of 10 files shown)


buunguyen:
Looks pretty a decent.

Some initial bugs:
* Should replace the toggle with a rectangular handle at the middle of the page. Also, replace cursor to normal instead of pointer, as it's no longer a button.
* The pin icon should change color when hover. 
* In float mode: click a file -> tree closes. I would expect it stay opens.
* In float mode: try to resize to make the tree wider -> close. Should stay open.
* In float mode: click a folder, then try to press any key -> close. Should stay open.
* In float mode: click a folder but outside the text (i.e. to the right of the text) -> close. Should stay open.
* In float mode: should slide in and out a bit smoother (and slower when hover/mouse out; okay to keep the same if pinned mode so that it doesn't delay the container push). Also, should wait 3-5secs with no interaction (unless users click outside) before closing the tree.
* In pin mode: the GitHub header isn't pushed, thus the tree hides GitHub logo and search bar.
* In general, the feeling using float mode is pretty weird right now (and quite buggy, e.g. sometime it closes when mouse is still in the tree, even click). Not super smooth and intuitive. Definitely need tweaking various things, e.g. speed, transition, behaviors... to make things better.

src/adapters/github.js
```diff
@@ -86,23 +86,14 @@ class GitHub extends PjaxAdapter {
...
-    const $header = $(GH_HEADER);
     const $containers = $(GH_CONTAINERS);
     const autoMarginLeft = ($(document).width() - $containers.width()) / 2;
-    const shouldPushEverything = sidebarVisible && autoMarginLeft <= sidebarWidth + SPACING;
+    const shouldPushEverything = sidebarPinned && sidebarVisible && autoMarginLeft <= sidebarWidth + SPACING;
 
     $('html').css('margin-left', shouldPushEverything ? sidebarWidth : '');
     $containers.css('margin-left', shouldPushEverything ? SPACING : '');
-
-    const headerPadding =
```
buunguyen: Remove this might be the cause of not pushing GitHub header.

src/adapters/github.less
```diff
@@ -174,6 +174,24 @@
...
+    & > i:before {
+      #octicons.pin();
+      color: @dark;
+      font-size: 15px;
+    }
+  }
+
+  a.octotree_pin-on > i:before {
+    #octicons.pin();
+    transform: rotate(-45deg);
```
buunguyen: Nice!

src/core.constants.js
```diff
@@ -1,24 +1,23 @@
 const NODE_PREFIX = 'octotree';
 const ADDON_CLASS = 'octotree';
 const SHOW_CLASS = 'octotree-show';
+const PINNED_CLASS = 'octotree_pin-on';
```
buunguyen: 'octotree_pinned'?

src/util.debounce.js
```diff
@@ -0,0 +1,20 @@
...
+      }
+      return
+    }
+
+    timeoutId = setTimeout(() => {
+      clearTimeout(timeoutId);
+      callback(...args);
+      timeoutId = null;
+    }, delay);
+  }
```
buunguyen: Missing semicolon. Recommend installing prettifier-atom if you're using Atom. Then choose "Format files on save".

src/util.debounce.js
```diff
@@ -0,0 +1,20 @@
...
+  let timeoutId = null;
+
+  return (...args) => {
+    timeoutId && clearTimeout(timeoutId);
+
+    if (cancelIf(...args)) {
+      if (timeoutId) {
+        timeoutId = null;
+      }
+      return
```
buunguyen: Semi.

src/util.debounce.js
```diff
@@ -0,0 +1,20 @@
+function debounce({callback, cancelIf, delay}) {
```
buunguyen: Please provide JSDoc comment for all new functions. Explain what each param is.

src/octotree.js
```diff
@@ -167,28 +182,36 @@ $(document).ready(() => {
...
-
     function toggleSidebar(visibility) {
       if (visibility !== undefined) {
         if (isSidebarVisible() === visibility) return;
         toggleSidebar();
       } else {
         $html.toggleClass(SHOW_CLASS);
         $document.trigger(EVENT.TOGGLE, isSidebarVisible());
+        if (isSidebarVisible()) {
+          tryLoadRepo();
```
buunguyen: Why reload repo here?

src/octotree.js
```diff
@@ -167,28 +182,36 @@ $(document).ready(() => {
...
+        }
+      }
+    }
+
+    function togglePinAndSave() {
+      store.set(STORE.PINNED, !isSidebarPinned(), togglePin);
+    }
+
+    function togglePin(isPinned) {
```
buunguyen: Is there a case when we toggle pin but doesn't save? If not, maybe move this logic inside togglePinAndSave (which can be rename to togglePin).

src/octotree.js
```diff
@@ -123,17 +138,17 @@ $(document).ready(() => {
...
       adapter.getRepoFromPath(currRepo, token, (err, repo) => {
         if (err) {
           showError(err);
         } else if (repo) {
           $toggler.show();
 
-          if (remember && shown) {
+          if (pinned) {
+            togglePin(true);
             toggleSidebar(true);
```
buunguyen: Maybe togglePin() should also show sideBar when true? Make the method simpler to use and reason about.

src/octotree.js
```diff
@@ -24,9 +25,23 @@ $(document).ready(() => {
...
+    }));
+
+    $toggler.mouseenter(() => toggleSidebar(true));
+    $toggler.click(() => toggleSidebar(false));
+
     key.filter = () => $toggler.is(':visible');
-    key(store.get(STORE.HOTKEYS), toggleSidebarAndSave);
+    key(store.get(STORE.HOTKEYS), () => {
+      togglePinAndSave();
+      toggleSidebar(isSidebarPinned());
```
buunguyen: See below. togglePin should do this automatically when 'true'.

src/octotree.js
```diff
@@ -24,9 +25,23 @@ $(document).ready(() => {
...
+      isSidebarPinned() && treeView.focus();
```
buunguyen: If togglePin return pin status, this could be simplify: if (togglePin()) treeView.focus();

src/octotree.js
```diff
@@ -24,9 +25,23 @@ $(document).ready(() => {
       if (event.target === window) layoutChanged();
     });
 
-    $toggler.click(toggleSidebarAndSave);
+    $pinner.click(togglePinAndSave);
+
+    $sidebar.on('keyup mouseleave mousemove', debounce({
+      callback: () => toggleSidebar(false),
+      cancelIf: ({type}) => isSidebarPinned() || type === 'mousemove',
+      deplay: DEFAULTS.SHOWTIME
```
buunguyen: typo: delay.

src/octotree.js
```diff
@@ -24,9 +25,23 @@ $(document).ready(() => {
       if (event.target === window) layoutChanged();
     });
 
-    $toggler.click(toggleSidebarAndSave);
+    $pinner.click(togglePinAndSave);
+
+    $sidebar.on('keyup mouseleave mousemove', debounce({
+      callback: () => toggleSidebar(false),
+      cancelIf: ({type}) => isSidebarPinned() || type === 'mousemove',
```
buunguyen: Avoid destructuring here. Change {type} to event, so it's clearer what this method takes in. (Not that I'm against destructuring, but this is a special case as debounce() return a fn(...args); so there must be a place that is explicit about the param type.)

buunguyen: || type === 'keyup'. Basically shouldn't close on if users still navigates with key.

src/octotree.js
```diff
@@ -24,9 +25,23 @@ $(document).ready(() => {
       if (event.target === window) layoutChanged();
     });
 
-    $toggler.click(toggleSidebarAndSave);
+    $pinner.click(togglePinAndSave);
+
+    $sidebar.on('keyup mouseleave mousemove', debounce({
```
buunguyen: I think this piece of logic is not right, causing the bugs I reported.

I think the logic could be like this:

* mouseleave: if (float mode) cause hiding timer to start (say, hide after 3s)
* mouseenter: remove any hiding timer
* keyup: reset any hiding timer (i.e. if keeps typing and there's a timer, restart it) 
* When toggle pin to true, remember to delete hiding timer.
* No need debouncing (unless mouseenter doesn't work in a certain edge case and we have to use mousemove; keyup shouldn't be a problem as it won't be that fast and a timer reset isn't too much work, prefer simpler code even if it's not as efficient, as long as it doesn't cause visible performance degrade).

src/octotree.js
nphi1212: The first load is necessary when opening the sidebar in the layout mode.

src/octotree.js
```diff
@@ -24,9 +25,26 @@ $(document).ready(() => {
...
+
+    $pinner.click(togglePin);
+
+    $sidebar.on(
+      `keyup mouseleave mousemove ${EVENT.PIN}`,
+      debounce({
+        callback: () => toggleSidebar(false),
+        // Mousemove handles the edge case in which mouseleave and then immediately mouseenter
+        cancelIf: (event) => isSidebarPinned() || event.type === 'mousemove',
+        delay: DEFAULTS.SHOWTIME
```
buunguyen: Shouldn't use DEFAULTS for non-storing var. Just introduce a constant, i.e. TREE_HIDING_DELAY

src/octotree.js
```diff
@@ -24,9 +25,26 @@ $(document).ready(() => {
...
-    $toggler.click(toggleSidebarAndSave);
+    $document.click((event) => !isSidebarPinned() && isOutsideSidebar(event.target) && toggleSidebar(false));
+
+    $pinner.click(togglePin);
+
+    $sidebar.on(
+      `keyup mouseleave mousemove ${EVENT.PIN}`,
+      debounce({
+        callback: () => toggleSidebar(false),
+        // Mousemove handles the edge case in which mouseleave and then immediately mouseenter
```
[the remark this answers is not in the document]
nphi1212: When user mouses out, it will trigger the timer. then user hovers back to the sidebar. We need to cancel the timer when user hovers back. debounce might be a misleading name. The purpose of listening the `mousemove` ('cancel' event) is cancelling timer if there is any. 

Beside "invoke after delay", the function also resets the timer in demand. I name the function `debounce` because it basically invokes the last `non-cancel` event. I realized the name is misleading. I will try to think another approach.

src/template.html
```diff
@@ -22,6 +20,10 @@
       <i class="settings"></i>
     </a>
 
+    <a class="octotree_spin">
+      <img src="https://assets-cdn.github.com/images/spinners/octocat-spinner-32-EAF2F5.gif" />
```
buunguyen: I don't think this is a good idea. We can't control it (e.g. what if the URL changes). Should stick with CSS-based spinner.

src/octotree.js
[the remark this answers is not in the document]
buunguyen: @nphi1212 any thought on my suggestion above?

src/adapters/github.js
```diff
@@ -89,8 +89,7 @@ class GitHub extends PjaxAdapter {
   updateLayout(sidebarPinned, sidebarVisible, sidebarWidth) {
     const SPACING = 10;
     const $containers = $(GH_CONTAINERS);
-    const autoMarginLeft = ($(document).width() - $containers.width()) / 2;
-    const shouldPushEverything = sidebarPinned && sidebarVisible && autoMarginLeft <= sidebarWidth + SPACING;
+    const shouldPushEverything = sidebarPinned && sidebarVisible;
```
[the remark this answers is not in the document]
buunguyen: Was this addressed?

src/octotree.js
```diff
@@ -167,28 +182,42 @@ $(document).ready(() => {
...
+    function togglePin(isPinned) {
+      if (isPinned !== undefined) {
+        if (isSidebarPinned() === isPinned) return;
+        return togglePin();
+      }
+
+      $pinner.toggleClass(PINNED_CLASS);
+      $document.trigger(EVENT.PIN, isSidebarPinned());
+
+      const sidebarPinned = isSidebarPinned();
```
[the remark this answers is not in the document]
nphi1212: I tried it but I think getting declaring the var after toggling  the `$spinner` makes the code cleaner. If declaring it at the top, we get the current state which hasn't toggled. As a result, the following logic will be reversed and hard to reason.

src/adapters/github.js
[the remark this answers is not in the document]
nphi1212: I have tried the approach with the old code, the logic is hard to reason about. I learned that there is a risk of removing the old code. Therefore, this time I have tested it more carefully.

src/octotree.js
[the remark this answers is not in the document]
nphi1212: @buunguyen  I agreed with you what it would be better to keep the logic explicitly in one place.  Will try to implement it =) Many thanks for your valued feedbacks

README.md
```diff
@@ -62,7 +62,6 @@ By default, Octotree only works on `github.com`. To support enterprise version (
```
crashbell: Please update the screenshots.

nphi1212: I don't have a big screen computer. Could someone help me with taking the screenshot ?

src/util.debounce.js
buunguyen: @nphi1212 whenever you've finished addressing a comment, please click "Resolve conversation" (if it also appears to you like it does to me).

buunguyen:
@nphi1212 	please re-issue PR to v3 branch after you've addressed the feedback. Thanks.

src/octotree.js
buunguyen: test

README.md
buunguyen: test

src/octotree.js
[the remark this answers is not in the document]
buunguyen: test

src/adapters/github.js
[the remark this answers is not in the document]
buunguyen: test

buunguyen:
test

README.md
buunguyen: test

src/octotree.js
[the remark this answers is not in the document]
buunguyen: test

src/adapters/github.js
buunguyen: test

src/adapters/github.js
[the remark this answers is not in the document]
buunguyen: new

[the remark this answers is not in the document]
buunguyen: new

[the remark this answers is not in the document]
buunguyen: new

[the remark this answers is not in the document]
buunguyen: new

README.md
buunguyen: new

src/adapters/github.js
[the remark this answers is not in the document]
buunguyen: new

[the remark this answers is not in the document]
buunguyen: new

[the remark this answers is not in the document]
buunguyen: new

PR #569, opened 2018-11-04
Show all files in PR view by recursively fetching all diffs. Fixes #504.

### LINKED ISSUE #504: Not showing all files in pull request view
### Description
Go to this PR : https://github.com/sil-project/Platform/pull/50/files

You can see more than 161 files.

The octotree plugin, that init the tree at page load, does not have the full file tree loaded, because Github load async the file list. So, as seen in the attached screenshot, the folder src/Sil/Bundle/ProductBundle/Resources/ is not shown

### Environment

* Octotree version: 2.4.6
* Browser & version: Firefox 60 Developer Edition
* OS & version: Ubuntu 16.04 up to date
* Screenshots :

stevennoto:
### Problem
The issue being tackled here is #504, i.e. Pull Request view not showing all files. This happens because the API only returns a certain number of files by default.

### Solution
Fortunately, the pull request API supports pagination, and we can use `page` and `per_ page` params to walk through the results. I moved some of the logic from `_getPatch` out to a new function `_getPatchRecursively` in order to repeatedly call the API until all results are fetched.

### Screenshots
This shows the new PR view. You can see the contents of `src/Sil/Bundle/ProductBundle/Resources/`, which as #504 mentions were not visible before. You can also see that more than 100 files (the old limit) were loaded into the tree (there are 190 under `src` alone).
octotree pr show all files

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -38,6 +38,7 @@ const GH_PJAX_CONTAINER_SEL = '#js-repo-pjax-container, .context-loader-containe
 const GH_CONTAINERS = '.container, .container-lg, .container-responsive';
 const GH_HEADER = '.js-header-wrapper > header';
 const GH_RAW_CONTENT = 'body > pre';
+const GH_PULLS_API_MAX_RESULTS = 100;
 
 class GitHub extends PjaxAdapter {
   constructor(store) {
@@ -211,17 +212,46 @@ class GitHub extends PjaxAdapter {
    */
   _getPatch(opts, cb) {
     const {pullNumber} = opts.repo;
+    const diffMap = {};
 
-    this._get(`/pulls/${pullNumber}/files?per_page=300`, opts, (err, res) => {
+    this._getPatchRecursively(pullNumber, 1, opts, diffMap, (err) => {
       if (err) cb(err);
       else {
-        const diffMap = {};
+        // Transform to emulate response from get `tree`
+        const tree = Object.keys(diffMap).map((fileName) => {
+          const patch = diffMap[fileName];
+          return {
+            patch,
+            path: fileName,
+            sha: patch.sha,
+            type: patch.type,
+            url: patch.blob_url
+          };
+        });
 
+        // Sort by path, needs to be alphabetical order (so parent folders come before children)
+        // Note: this is still part of the above transform to mimic the behavior of get tree
+        tree.sort((a, b) => a.path.localeCompare(b.path));
+
+        cb(null, tree);
+      }
+    });
+  }
+
+  _getPatchRecursively(pullNumber, pageNumber, opts, diffMap, cb) {
+    const url = `/pulls/${pullNumber}/files?per_page=${GH_PULLS_API_MAX_RESULTS}&page=${pageNumber}`;
+    this._get(url, opts, (err, res) => {
+      if (err) cb(err);
+      else {
+        let numResults = 0;
+        let offset = GH_PULLS_API_MAX_RESULTS * (pageNumber - 1);
         res.forEach((file, index) => {
+          numResults++;
+
           // Record file patch info
           diffMap[file.filename] = {
             type: 'blob',
-            diffId: index,
+            diffId: index + offset,
             action: file.status,
             additions: file.additions,
             blob_url: file.blob_url,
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -260,23 +290,11 @@ class GitHub extends PjaxAdapter {
...
-          };
-        });
-
-        // Sort by path, needs to be alphabetical order (so parent folders come before children)
-        // Note: this is still part of the above transform to mimic the behavior of get tree
-        tree.sort((a, b) => a.path.localeCompare(b.path));
-
-        cb(null, tree);
+        if (numResults === GH_PULLS_API_MAX_RESULTS) {
+          this._getPatchRecursively(pullNumber, pageNumber + 1, opts, diffMap, cb);
```
duylam: Hey @stevennoto

Thanks for great work. I've read through code and I'd like to share below 02 concerns on the approach

1. By sending multiple requests per PR detail page, there is a risk of the viewing PR action may kill the Github API rate for unauthenticated user. See at https://developer.github.com/v3/#rate-limiting. 

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address

2. By calling the fetching file list recursively, it may take long time for users to see the file list in sidebar while the Github already displays them. This somehow makes user confused and inconvenient 

What do you think ?

buunguyen: Concur with @duylam.

For rate limit, maybe we should only do this when user supplied access token? Should change the hint text to be clear (e.g. "Only limit to 100 files if no access token due to [rate limit](link)").

For performance, that is certainly a concern. For large repos, the practice is people would uncheck "Load entire tree at once", so Octotree would lazy load directories. But for this I'm not sure how to best deal with huge PRs. Ideas are welcome.

stevennoto: Thanks for the excellent feedback!

Regarding the rate limit, it would be simplest to only fetch more pages when the user has supplied an access token. Or we could add an additional option under "Show only pull request changes" like "Fetch all files for large pull requests" to let users control this independently.

Regarding performance, I am not sure. We could look at the user's choice for "Load entire tree at once," but the idea of lazy loading does not work as well in PR view, since the first 100 files we get are not necessarily all the root items, and there would be no indication which ones are missing, to give the user something to click on to lazily-load the rest.

One idea would be to draw the tree after each API call, gradually filling in all the items. However I think this would significantly complicate the callback code. And I am not sure if the tree can remember which items a user has expanded when it redraws.

Another idea would be to fetch the first 100 items only, but add an indication somewhere that more might exist, and show a "Load More" button to continue fetching. Another idea would be the opposite, to show a label like "Loaded 200 items, loading 201-300" etc, and provide a button to "Stop Loading." These feel very manual to me though.

Since this would be an issue mainly on very large PR's, like 500+ files, maybe adding an option like "Fetch all files for large pull requests" would be simplest.

duylam: @stevennoto

Thanks for giving many alternative solution on the issues

> Regarding the rate limit, it would be simplest to only fetch more pages when the user has supplied an access token

I'd love this solution more than a toggle option for "fetching all files". A problem with option "fetching all files" is that Octotree may stop working due to out of api call quota (e.g. stop working) even users may aware about this risk. And that risk is more critical to me than listing less files in PR

> to give the user something to click on to lazily-load the rest.

From options for addressing the performance, I prefer above suggestion. IMO, users should proactively indicate loading all files in PR whenever they need. By manually telling Octotree to load all files, they can wait for a while to see all files. Any effort for automating loading all files would not help in this case

For the UI, I think Octotree might display an icon next to the gear (setting) icon for trigger loading all files in PR. But this icon should display only in PR page. I know it doesn't look perfect but I don't have better idea

buunguyen: For now, my vote would be on the simplest approach that cover the most use cases, but not hurting rate limit. I would say:
* No token: 100 files
* With token: 500 files

We should update the hint text underneath the show-only-changes-in-pr option to make that clear.

isbadawi: An alternative, instead of using that endpoint (which according to the docs is limited to 300 files), could be to download the diff by doing a GET on the regular PR url but with a `Accept: application/vnd.github.v3.diff` header, then look for all the file headers. I think this will get you everything in one request (though I haven't tested with super huge PRs). However, the response could be very large.

isbadawi: I just verified on a PR with 324 modified files that I could only get at the first 300 by paging through the `/files` endpoint, but I was able to get all of them by downloading the diff and parsing it (actually running `git apply --numstat`). And the response isn't that much larger -- I didn't realize earlier that the `/files` endpoint also gives you the patch contents.

isbadawi: I also tried using this API to compare the PR base sha and head sha: https://developer.github.com/v3/repos/commits/#compare-two-commits
But it also only returned 300 files.

buunguyen: @isbadawi thanks for your research. Get only diff sounds great if the response is indeed not much bigger than the 300 files (with patched contents).
@stevennoto what do you think?

BlythMeister:
This pull request has been sitting here for a while.  what needs to be done to get this issue resolved

buunguyen:
This PR is obsolete now. We'll figure something out.

PR #573, opened 2018-11-15
Auto save options in the Option view

duylam:
### Solution

Enhance the Option view: no need the Save button

### DIFF
--- a/src/view.options.js
+++ b/src/view.options.js
@@ -1,9 +1,11 @@
 class OptionsView {
   constructor($dom, store) {
+    this.$this = $(this);
     this.store = store;
-    this.$view = $dom.find('.octotree_optsview').submit(this._save.bind(this));
+    this.$view = $dom.find('.octotree_optsview');
     this.$toggler = $dom.find('.octotree_opts').click(this._toggle.bind(this));
     this.elements = this.$view.find('[data-store]').toArray();
+    this.optionChanges = {};
 
     // Hide options view when sidebar is hidden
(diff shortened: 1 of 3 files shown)


buunguyen:
@duylam please re-issue this to v3 branch.

src/view.options.js
```diff
@@ -28,65 +35,78 @@ class OptionsView {
...
-        if ($elm.is(':checkbox')) $elm.prop('checked', value);
+        const checkboxElement = $elm.is(':checkbox')
+
+        if (checkboxElement) $elm.prop('checked', value);
         else $elm.val(value);
+
+        // Attaches (once) on value change event
+        if (!$elm.data('octotree-on-change-event-attached')) {
+          $elm.data('octotree-on-change-event-attached', true);
+          const boundOnChange = this._saveOption.bind(this, $elm, key);
```
buunguyen: Why not save only when users switch to a different view?
Would be simpler in logic. And reuse current saveOptions().

src/view.options.js
```diff
@@ -28,65 +35,78 @@ class OptionsView {
...
     this._eachOption(
       ($elm, key, value, cb) => {
-        if ($elm.is(':checkbox')) $elm.prop('checked', value);
+        const checkboxElement = $elm.is(':checkbox')
+
+        if (checkboxElement) $elm.prop('checked', value);
         else $elm.val(value);
+
+        // Attaches (once) on value change event
+        if (!$elm.data('octotree-on-change-event-attached')) {
```
buunguyen: Why is this necessary?

PR #577, opened 2018-11-20
Add Octotree service to centralize some common tasks

duylam:


gulpfile.js
```diff
@@ -181,6 +181,7 @@ function buildJs(prefix = '.', ctx = {}) {
     `${prefix}/src/core.constants.js`,
     `${prefix}/src/core.storage.js`,
     `${prefix}/src/core.plugins.js`,
+    `${prefix}/src/core.octotree.js`,
```
buunguyen: I think we should rename this to something else. Maybe core.api?

src/adapters/adapter.js
```diff
@@ -193,11 +193,11 @@ class Adapter {
...
           'and uncheck the "Load entire tree at once" option.';
         break;
+
+      // Errors for access token
       case 401:
         error = 'Invalid token';
-        message =
-          'The GitHub access token is invalid. ' +
-          'Please go to <a class="settings-btn" href="javascript:void(0)">Settings</a> and update the token.';
+        message = window.octotreeService.getAccessTokenErrorMessage(settings, jqXHR);
```
buunguyen: Rename to just `octotree`? Also, no need `window`.

src/core.octotree.js
```diff
@@ -0,0 +1,21 @@
+class OctotreeService {
+  getAccessToken() {
+    return window.store.get(window.STORE.TOKEN);
+  }
+
+  getAccessTokenErrorMessage(settings, jqXHR) {
+    let message = '';
+
+    switch (jqXHR.status) {
```
buunguyen: 1. Rename to getInvalidTokenMessage.
2. Don't pass jqXHR -> the caller only calls this in 401, so need to check and switch again. Don't pass the entire request settings, just the properties you need.

src/core.plugins.js
```diff
@@ -63,6 +63,23 @@ class Plugin {
     return undefined;
   }
 
+  /**
+   * Initializes the plugin.
```
buunguyen: Comments on the difference between activate and initialize. Why can't use just one? Or why not put this in ctor of plugin?

src/main.js
```diff
@@ -112,6 +115,7 @@ $(document).ready(() => {
     function tryLoadRepo(reload) {
       const token = store.get(STORE.TOKEN);
```
buunguyen: Remove this line.

duylam:
@buunguyen Some refactoring at https://github.com/ovity/octotree/pull/577/commits/528d7f5547c40bc243c829bcf53af390c561fa76

PR #578, opened 2018-11-22
Remove the constant background color.

xingheng:
It uses the hard background color for the bottom div layer and looks ugly in the dark mode when applying the custom style from stylish extension in google chrome.


buunguyen:
Please consider Octotree Pro if you want built-in dark mode support: https://www.octotree.io/.

Issue #596, opened 2019-01-09
Option to disable sidebar and folder animations


RomanKornev:
### Description
Sidebar sliding animation can be extremely annoying when working with multiple projects.
Disabling it helps dramatically.
Also disabling folder opening/closing animations can speed up workflow on deep repositories.

I'll go ahead and submit a PR.

buunguyen:
We reduced animation 4x. It's not too abrupt but too slow. Hope it's the right balance. Gonna close this issue now.

PR #597, opened 2019-01-09
Add Enable/Disable animations option

RomanKornev:
### Problem
For #596 

### Solution
Adds a new option Enable animations.
It's checked by default - nothing has changed.
If it's unchecked - disables sidebar transition and folder open/close animations.
Default `jstree` animation time is 200ms.

#### Potential problem: 
Page needs to be reloaded after changing the setting to see the effect.
If anyone can suggest a fix, I would appreciate it because I'm not a JS dev.

### Screenshots
disable-animations


duylam:
@RomanKornev Some general feedbacks:

- Can you look at the conflict file?
- The animation is controlled in css, why don't you go with changing css (see https://github.com/ovity/octotree/blob/v3/src/adapters/github.less#L13) ?

PR #599, opened 2019-01-10
Downgrade event-stream to non-compromised version.

casper6863:
### Problem
Describe the problem.

### Solution
Describe the solution.

### Screenshots


buunguyen:
It was done last month I think: https://github.com/ovity/octotree/commit/27c346017ee6024895a9312985102fa28616a4b5

PR #614, opened 2019-01-25
Fix issue of folders opening or closing twice in a row

### LINKED ISSUE #613: Folders closing or opening twice in a row
### Description
Very often when I attempt to close or open a folder in the tree view, the following can happen:
- If the folder is closed, it will stay closed
- If the folder is opened, it will transition instantly to the close state and redo the opening animation 
- Usually it can be reproduced by clicking one folder, then clicking another different folder twice

### Environment (if bug)

* Octotree version: 2.5.1
* Browser & version: Firefox 64.0.2 and Chrome 71.0.3578.98 (But it doesn't seem to be browser-specific)
* OS & version: MacOS Sierra 10.12.6

Sayegh7:
### Problem
Very often when I attempt to close or open a folder in the tree view, the following can happen:
- If the folder is closed, it will stay closed
- If the folder is opened, it will transition instantly to the close state and redo the opening animation 
- Usually it can be reproduced by clicking one folder, then clicking another different folder twice

### Solution
By wrapping the close and open functions in a setTimeout, the issue disappears. It seems like the issue happens due to the functions getting called too early.

### Screenshots
N/A

closes #613

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -5,8 +5,16 @@ class TreeView {
     this.$view = $dom.find('.octotree_treeview');
     this.$tree = this.$view
       .find('.octotree_view_body')
-      .on('click.jstree', '.jstree-open>a', ({target}) => this.$jstree.close_node(target))
-      .on('click.jstree', '.jstree-closed>a', ({target}) => this.$jstree.open_node(target))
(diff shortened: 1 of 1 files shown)


buunguyen:
Great fix, thank you!

Issue #490, opened 2018-02-07
`pjax:end` event not dispatched on native `document`


josephfrazier:
### Description
Over at https://github.com/OctoLinker/OctoLinker/issues/416, we're having a compatibility bug between Octotree and Octolinker. https://github.com/OctoLinker/OctoLinker/issues/416#issuecomment-363576288 explains:

> I found the reason why this is not working. Whenever you navigate between files, PRs or issues on GitHub, they fetch the HTML from the server via ajax and replaces the content of container with the loaded html. An event from type `pjax:end` is dispatched on `document` when this occurs. 
> 
> Octotree, is using basically the same technique to load the pages with a small difference. Other than GitHub.com they use a [pjax jQuery plugin](https://github.com/defunkt/jquery-pjax) to achieve this. Since this is a jQuery plugin they dispatch all events on a wrapped `document` which leads to jQuery events rather than native events. By the nature of chrome extensions, Octotree and OctoLinker have different jQuery instances which makes in impossible to listen for events from the other extension.
> 
> Currently there are only two feasible options to solve this issue. Either creating an Octotree PR to dispatch also a native event on `document` or we revert https://github.com/OctoLinker/injection/pull/12/commits/0dcf827eb6ca902139c9384f54b139ee656246f3 which was moving away from listening for dom changes in favor of `pjax:end`. The latter feels wrong to me, since other extensions could still have the same issue if they also rely on the `pjax:end` event

I'm not too familiar with Octotree personally, but does this sound accurate to you? I think the suggested fix would just be a `document.dispatchEvent` call added [here](https://github.com/buunguyen/octotree/blob/bd8122dd8e3c7757149bd44b0899733450942b11/src/adapters/pjax.js#L8), let me know if that seems right.

### Environment

* Octotree version: 2.4.3
* Browser & version: Firefox/Chrome 57.0/63.0.3239.84
* OS & version: N/A
* Screenshot, if any:

![](https://user-images.githubusercontent.com/6207220/34162396-d76a1476-e4d3-11e7-82dd-e4718d218901.gif)

sirkuttin:
this bug is the only thing in the way from my github workflow from being perfect

kidonng:
> The latter feels wrong to me, since other extensions could still have the same issue if they also rely on the pjax:end event

So this is happening to Refined GitHub sindresorhus/refined-github#1759.
@buunguyen Could you please take a look at this? Most people use both the extensions and this could be a frequent issue.

[...]

PR #616, opened 2019-01-27
Dispatch 'pjax:end' event on native document

josephfrazier:
This attempts to address https://github.com/ovity/octotree/issues/490,
but I haven't yet tested it myself.


buunguyen:
Thanks for this, but gonna close until we know of a solution that is verified to be working.

PR #618, opened 2019-01-29
Private repository load error in private repository

gongul:
### Problem
Press the toggle button twice.  -> Press the setting button twice. -> An error occurs.

### Solution
src / octotree.js 254 
`hasError = false;` remove 

### Screenshots
same repository 

[before google drive](https://drive.google.com/open?id=1uH-W3a7gq-bqZFYQ21Y-5Ha6Q2QfJ3EO)
[after google drive](https://drive.google.com/open?id=1uRbU0wYVtpDGXSLEmN8BkX77aLuAbk2E)

### DIFF
--- a/src/octotree.js
+++ b/src/octotree.js
@@ -122,7 +122,6 @@ $(document).ready(() => {
     }
 
     function tryLoadRepo(reload) {
-      hasError = false;
       const remember = store.get(STORE.REMEMBER);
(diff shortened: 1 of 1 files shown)


buunguyen:
I can't reproduce this error in v3 branch, which would be the new release. For the current version, that flag is there for a reason, removing it would cause another problem.

Issue #539, opened 2018-09-18
Safari latest update


vigo:
### Description
After todays update on Safari, octotree is gone :(

### Environment (if bug)

* Octotree version:
* Browser & version: Version 12.0 (13606.2.11)
* OS & version: OSX 10.13.6 (17G65)
![screen shot 2018-09-18 at 17 11 41](https://user-images.githubusercontent.com/82952/45693511-f0275280-bb65-11e8-85a6-de078c832fec.png)

* Screenshot, if any (drag an image here)

Please help! :)

vigo:
@ravicious well, maybe we "Safari" users should start a mini campaign? If we find 10 people, $99/10 less than $10 for each person? Who wants to help? ✋

darrenchaker:
Thank you for the update! Very simple to add even for novice user - Best, Darren Chaker

tbodt:
I have an apple developer program membership, and I'd be willing to volunteer that to put Octotree in the app store. @buunguyen

sandangel:
any update on this? Really need this to use safari, chrome and firefox use too much memory on my machine

buunguyen:
@tbodt thanks. I'll review safari support and let you know if I need help publishing.

tomisacat:
Any update?

thesunwave:
Is there any update?

loveforgeter:
any update😞?

rozag:
> @tbodt thanks. I'll review safari support and let you know if I need help publishing.

Is there any update on that?

IvanMwiruki:
@buunguyen Any updates? Would really love to have safari support back.

linjson:
@buunguyen when do you support on **safari 12**!!!!!!

andreasowczarek:
You might find this one helpful https://georgegarside.com/blog/macos/install-any-safari-extension-macos-mojave/

linjson:
@andreasowczarek, very good!!

buunguyen:
I'm excited to announce that Octotree Pro is now an official Safari App Extension: https://itunes.apple.com/us/app/octotree-pro/id1457450145?mt=12.

This is a premium version of Octotree that has advanced features such as dark mode, file search etc. You can learn more about Octotree pro here: https://www.octotree.io/.

[...]

PR #619, opened 2019-01-30
Add instructions for building for safari via docker

pmellati:
### Problem
The existing documentation for safari builds are hard to use, and the prebuilt safari package is not updated.

### Solution
Add instructions for building for safari via a docker container.

### Screenshots

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Browser extension (Chrome, Firefox, Opera and Safari) to show a code tree on Git
 
 __Note__: to configure Octotree for GitHub Enterprise, see [instructions](#enterprise-urls)
 
-### Install on Safari
+### Install on Safari (without using docker)
 
 Octotree is not available on the Safari gallery. You have to build from source.  Download this repo and run the following command:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,43 @@ gulp safari
...
+First, run interactive bash in a container that has all the required tools:
+
+```bash
+docker run -it --name octotree-build gabrielaraujof/nodejs-gulp-bower bash
+```
+
+Within the container:
+
+```bash
+git clone https://github.com/ovity/octotree.git
```
ghostiam: I suggest adding `--depth 1`, so cloning happens faster, and we do not need the whole history of the repository

README.md
```diff
@@ -28,6 +28,43 @@ gulp safari
...
+cd octotree/
+npm install
+npm install gulp
+gulp safari
```
ghostiam: First need to build styles, and only then the extension for safari

```suggestion
gulp styles
gulp safari
```

buunguyen:
https://github.com/ovity/octotree/issues/539#issuecomment-476486281

Issue #595, opened 2019-01-05
Switch to lazy loading when "Load entire tree at once" is required.


samvel1024:
Sometimes when I browse a huge repo, the tree is not rendered and a message asks to enable "Load entire tree at once". ([example](https://github.com/Stichting-MINIX-Research-Foundation/minix)) However when I enable it I get an annoying lag when I browse small repositories since the implementation of the lazy strategy is fetching each directory separately. 

My suggestion is to dynamically switch the strategy from eager to lazy when the user tries to browse a huge repository, because having "Load entire tree at once" enabled all the time slows down the user a little bit. I would be happy to help in the realisation of this feature.

Thanks for the great utility

buunguyen:
Sounds reasonable. Maybe handle that error, then automatically lazy for that repo only. Can you send a PR?

buunguyen:
Closed as it was merged.

PR #620, opened 2019-02-07
Mark huge repositories to load lazily later (#595)

samvel1024:
When it discovers that repo is too big to load it marks it and stores in the storage. Before each request it checks if the repo is huge. 

The first time it wants to fetch a huge repo it asks to refresh the page. This is "hack"-ish fix, since we need to retry loading lazily immediately. Could not find an elegant way to do this.

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -237,7 +237,8 @@ class Adapter {
     cb({
       error: `Error: ${error}`,
       message: message,
-      needAuth: needAuth
+      needAuth: needAuth,
+      status: jqXHR.status
     });
   }
 
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -191,8 +191,7 @@ class Adapter {
       case 206:
         error = 'Repo too large';
         message = `This repository is too large to loaded in a single request.
-          If you frequently work with this repository,
-          go to Settings and uncheck the "Load entire tree at once" option.`;
+          For subsequent requests the repository will be loaded lazily. Please refresh the page to see the results.`;
         break;
       case 401:
         error = 'Invalid token';
@@ -271,7 +270,7 @@ class Adapter {
    * a single request. This is usually determined by the underlying the API.
    * @api public
    */
-  canLoadEntireTree() {
+  canLoadEntireTree(opts) {
     return false;
   }
 
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -188,11 +188,6 @@ class Adapter {
           Please try again later.`;
         needAuth = false;
         break;
-      case 206:
-        error = 'Repo too large';
-        message = `This repository is too large to loaded in a single request.
-          For subsequent requests the repository will be loaded lazily. Please refresh the page to see the results.`;
-        break;
       case 401:
         error = 'Invalid token';
         message = `The token is invalid.
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -38,6 +38,7 @@ const GH_PJAX_CONTAINER_SEL = '#js-repo-pjax-container, .context-loader-containe
 const GH_CONTAINERS = '.container, .container-lg, .container-responsive';
 const GH_HEADER = '.js-header-wrapper > header';
 const GH_RAW_CONTENT = 'body > pre';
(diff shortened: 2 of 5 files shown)


buunguyen:
Thanks for working on this. I'd prefer a more transparent approach where users don't have to do anything. Likely have to catch the error early and kick in the reload programmatically.

samvel1024:
Ok, I have thought of that, but tried to do in a more elegant way using callbacks. I will change it to the way you suggest (using exceptions) in the next commit.

samvel1024:
I have modified it so that we catch the error and retry the loading programatically.

src/view.tree.js
```diff
@@ -23,13 +23,17 @@ class TreeView {
...
-      const loadAll = this.adapter.canLoadEntireTree() && (prMode || this.store.get(STORE.LOADALL));
+      const loadAll = this.adapter.canLoadEntireTree(repo) && (prMode || this.store.get(STORE.LOADALL));
 
       node = !loadAll && (node.id === '#' ? {path: ''} : node.original);
 
       this.adapter.loadCodeTree({repo, token, node}, (err, treeData) => {
         if (err) {
-          $(this).trigger(EVENT.FETCH_ERROR, [err]);
+          if (err.status === 206 && loadAll) { // The repo is too big to load all, need to retry
+            $jstree.refresh(true);
```
buunguyen: What does this do? Refresh lazily?

src/adapters/github.js
```diff
@@ -306,6 +308,12 @@ class GitHub extends PjaxAdapter {
     $.ajax(cfg)
       .done((data) => {
         if (path && path.indexOf('/git/trees') === 0 && data.truncated) {
+          const hugeRepos = this.store.get(STORE.HUGE_REPOS);
+          const repo = `${opts.repo.username}/${opts.repo.reponame}`;
+          if (undefined === hugeRepos[repo]) {
+            hugeRepos[repo] = true;
```
buunguyen: My concern is this list will keep growing and local storage is finite. At some point (probably a long time, but still..), Octotree won't be able to save anything any more. Maybe should keep only the most 100 recent large repos?

src/adapters/github.js
```diff
@@ -306,6 +308,12 @@ class GitHub extends PjaxAdapter {
     $.ajax(cfg)
       .done((data) => {
         if (path && path.indexOf('/git/trees') === 0 && data.truncated) {
+          const hugeRepos = this.store.get(STORE.HUGE_REPOS);
+          const repo = `${opts.repo.username}/${opts.repo.reponame}`;
+          if (undefined === hugeRepos[repo]) {
```
buunguyen: `if (!hugeRepo[repo])`

src/adapters/github.js
```diff
@@ -73,8 +73,10 @@ class GitHub extends PjaxAdapter {
   }
 
   // @override
-  canLoadEntireTree() {
-    return true;
+  canLoadEntireTree(repo) {
+    const key = `${repo.username}/${repo.reponame}`;
+    const hugeRepos = this.store.get(STORE.HUGE_REPOS);
+    return !(hugeRepos[key] === true);
```
buunguyen: `return !hugeRepos[key];`

src/view.tree.js
samvel1024: It checks wether the tree was too big , if yes it performs a reload instead of showing an error message. Since the number 206 refers to the HTTP status code, I felt like it will be odd to name it somehow through a constant. But still I agree that it is kind of ugly.

src/adapters/github.js
samvel1024: Good point, I'll handle that

samvel1024:
It turns out it's difficult to find a big enough repo to test the extension 😄I have generated two pretty big repositories for testing. 

https://github.com/samvel1024/deeeep_tree1
https://github.com/samvel1024/deeeep_tree2

src/adapters/adapter.js
```diff
@@ -191,8 +191,7 @@ class Adapter {
       case 206:
```
buunguyen: Now that we handle 206, I suppose we should remove this error handling here? Sorry I missed it in the earlier review.

samvel1024: My bad. Done.

buunguyen:
Great work! Thanks for working on this.

PR #626, opened 2019-02-17
Fix(branch name guessing): handle `/` in branch name when using link

enizor:
### Problem

The branch name is not correctly parsed when it contains a `/`

### Solution

When parsing the commits list, delete the initial `/username/reponame/commits/` instead of getting the last `/`.

You can check with this [PR branch](https://github.com/enizor/octotree/tree/enizor/fix-branch): the extension tried to use `fix-branch` instead of the whole `enizor/fix-branch` 

Maybe `currentRepo.username` should be used instead of only `username` (and similarly for `reponame`) but I did not understand when the two could differ

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -170,9 +170,9 @@ class GitHub extends PjaxAdapter {
       (type === 'commit' && typeId) ||
       // Pick the commit ID or branch name from the DOM
       // Note: we can't use URL as it would not work with branches with slashes, e.g. features/hotfix-1
+      branchFromSummary ||
       ($('.overall-summary .numbers-summary .commits a').attr('href') || '')
         .replace(`/${username}/${reponame}/commits/`, '') ||
-      branchFromSummary ||
       // Pull requests page
       ($('.commit-ref.base-ref').attr('title') || ':').match(/:(.*)/)[1] ||
       // Reuse last selected branch if exist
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -170,7 +170,8 @@ class GitHub extends PjaxAdapter {
       (type === 'commit' && typeId) ||
       // Pick the commit ID or branch name from the DOM
       // Note: we can't use URL as it would not work with branches with slashes, e.g. features/hotfix-1
-      ($('.overall-summary .numbers-summary .commits a').attr('href') || '').split('/').slice(-1)[0] ||
+      ($('.overall-summary .numbers-summary .commits a').attr('href') || '')
+        .replace(`/${username}/${reponame}/commits/`, '') ||
```
buunguyen: Maybe we should move this line after line 175 (`branchFromSummary ||`).

enizor: That's a good idea (and fixes the issue by itself).

buunguyen: Yeah, this is the fallback. The other one is more robust. Thus still need this line here though in case DOM changes.

enizor:
On a side note, i needed to install `natives@1.1.6` to build (`ReferenceError: internalBinding is not defined`). Is it a problem on my side?

PR #629, opened 2019-02-21
Fix: fix error header format

BigTony666:
### Problem
Error header does not have correct message format, this PR is aimed to fix the format style.

### Solution
Modify the `github.less` file.

### Screenshots

Original:
original

Modified:
modified


buunguyen:
This was already fixed in another branch. To be released soon.

PR #630, opened 2019-03-03
Prevent mouseenter triggered when mouse is out of the sidebar

nphi1212:
### Problem
Click file and mouse out doesn't automatically hide the sidebar in float mode. The reason is when loading a new file, `mouseenter` in the side is triggered after the file is loaded even the mouse is moved out the sidebar already.
### Solution
Check if mouse is out the sidebar in the `mouseenter` event.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -233,7 +233,12 @@ $(document).ready(() => {
         startTimer(MOUSE_LEAVE_DELAY);
       });
       $sidebar.on('mouseenter mousemove', (event) => {
-        // Mouse is out
+        /**
+         * When loading a new file, the page is re-rendered,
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -232,7 +232,10 @@ $(document).ready(() => {
         isMouseInSidebar = false;
         startTimer(MOUSE_LEAVE_DELAY);
       });
-      $sidebar.on('mouseenter mousemove', () => {
+      $sidebar.on('mouseenter mousemove', (event) => {
+        // Mouse is out
```
buunguyen: Can you add more detailed comment here and why does this fixes it. I'm sure 3 months later when we look back, we'll be surprised what this is about.

PR #633, opened 2019-03-05
Bug: double click on the folder lead to hiding the sidebar in floatmode

nphi1212:
### Problem
When double click on any folder would hide the sidebar in the float mode.
### Solution
Ensure click event (which immediately hides the sidebar) being triggered outside the sidebar

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -206,7 +206,7 @@ $(document).ready(() => {
     function setupSidebarFloatingBehaviors() {
       const MOUSE_LEAVE_DELAY = 2000;
       const KEY_PRESS_DELAY = 4000;
-      let isMouseInSidebar = false;
+      const notInSidebar = (event) => !$sidebar.find(event.target).length;
 
       // Opens when mouse is over the handle.
       $toggler.mouseenter(() => toggleSidebar(true));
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -212,8 +212,13 @@ $(document).ready(() => {
       $toggler.mouseenter(() => toggleSidebar(true));
 
       // Immediately closes if click outside the sidebar.
-      $document.on('click', () => {
-        if (!isMouseInSidebar && !isSidebarPinned() && isSidebarVisible()) {
+      $document.on('click', (event) => {
+        /**
+         * Ensure the event click being triggered outside the sidebar
+         */
+        const notInSidebar = !$sidebar.find(event.target).length;
```
buunguyen: So we're not using the var 'isMouseInSidebar' anymore. Options:

1. Should remove it.
2. Should fix the code so that isMouseInSidebar has the right value, so that doesn't have to change this piece

buunguyen:
@nphi1212 for some reason I can't reproduce this bug on my machine (Chrome/OSX).

buunguyen:
@nphi1212 	okay, can reproduce now right after I said that. Need to click near the edge, outside the folder text.

src/main.js
```diff
@@ -206,32 +206,32 @@ $(document).ready(() => {
...
         }
       };
       const clearTimer = () => timerId && clearTimeout(timerId);
 
-      $sidebar.on('keyup', () => startTimer(KEY_PRESS_DELAY));
-      $sidebar.on('mouseleave', () => {
-        isMouseInSidebar = false;
-        startTimer(MOUSE_LEAVE_DELAY);
-      });
+      $sidebar.on('keyup', () => startTimer(KEY_PRESS_DELAY, false));
```
buunguyen: Hmm, this doesn't make sense. So it always passes 'false', meaning the timer doesn't start. So what's he point of this entire inline at all?

src/main.js
```diff
@@ -206,32 +206,32 @@ $(document).ready(() => {
     function setupSidebarFloatingBehaviors() {
       const MOUSE_LEAVE_DELAY = 2000;
       const KEY_PRESS_DELAY = 4000;
-      let isMouseInSidebar = false;
+      const notInSidebar = (event) => !$sidebar.find(event.target).length;
```
buunguyen: Should be called isInsideSidebar(). Don't do negative.

Issue #594, opened 2019-01-02
Footer covers last item


benwaffle:
### Description
Missing `dev.go`:

<img width="446" alt="image" src="https://user-images.githubusercontent.com/1713819/50616118-28546600-0eb5-11e9-9451-4be2c7f17f3c.png">

### Environment (if bug)

* Octotree version: 2.5.1
* Browser & version: Firefox 64.0
* OS & version: macOS 10.14.2

buunguyen:
Fixed in v3, will release soon.

PR #637, opened 2019-03-14
Footer covering the last item fixed

DhvanilP:
### Problem
#594 Footer cover the last item

### Solution
Leave margin at bottom so that last file/folder is visible in the tree.

#### Before fix
Screenshot from 2019-03-14 14-22-00

#### After fix
Screenshot from 2019-03-14 14-21-31

### DIFF
--- a/src/styles/base.less
+++ b/src/styles/base.less
@@ -58,6 +58,7 @@
 
   .octotree_view {
     display: none;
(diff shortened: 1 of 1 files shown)


buunguyen:
Thanks but this was already fixed in v3 branch, which will be released soon.

PR #645, opened 2019-04-07
adjust content position based on header height

nphi1212:
Screen Shot 2019-04-07 at 1 26 41 PM
Screen Shot 2019-04-07 at 1 27 02 PM
Screen Shot 2019-04-07 at 1 27 59 PM

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -78,6 +78,8 @@
 
       .octotree-view-header {
         position: absolute;
+        display: flex;
+        align-items: center;
         background-color: @dark-less;
         border-right: 1px @dark solid;
         color: white;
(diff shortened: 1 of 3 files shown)


buunguyen:
Thanks for this. The icons don't look super aligned, i guess due to the
shape, but can you check and make it look more aligned, maybe position
relative and move up or down a bit. If you don't see the problem I can
merge and fix myself

nphi1212:
The rotated pin icon is bigger than the idle one. I have fixed it by adding `scale(0.9)` when it's rotated.

dinkelaker:
Sorry, I approved the review by mistake (clicked the wrong button) ;-)

Issue #648, opened 2019-04-13
Disable drawer open on hover


SpencerPark:
This is more of a feature request but really is a request for reverting a "feature". 

octotree just updated on me to v3.0.0 and the new sidebar (which I keep closed unless looking for files) has the new functionality of opening on mouseover and it is getting a little frustrating. Could the hover functionality be under a flag instead? I would only like the drawer to open when I click on the button.

Otherwise the drawer keeps opening when I just want to move my mouse out of the way, or in another case where my second monitor is to the left and my mouse crosses over the boundary to get there, again the drawer opens when I wouldn't like it to.

Thanks for all your work on octotree, it makes github much more enjoyable :)

Dekken:
Really not liking the new update either - this is one reason

buunguyen:
Providing an option to click to open makes sense. We'll work on that.

buunguyen:
@Dekken 

> Really not liking the new update either - this is one reason

What are the other reasons?

hsluoyz:
The sidebar keeps opening by itself all the time when I looked at a GitHub repo. I still prefer the original button. Is there any way to go back? Thanks.

nphi:
I have created a PR for switching between hover and click behaviour. I hope it would be merged ASAP

laudavid:
I do not like the feture too, please fix it as soon as possible .

nphi:
@hsluoyz  I guessed you're in the pinned mode by using hotkeys. Could you check if the pin icon is on?
![Screen Shot 2019-04-14 at 3 32 28 PM](https://user-images.githubusercontent.com/8633840/56090294-b2f02c80-5eca-11e9-810b-2c97a2faf509.png)

Dekken:
> @Dekken
> 
> > Really not liking the new update either - this is one reason
> 
> What are the other reasons?

It has disabled use of user styles (stylus/stylish)

tamer1an:
@Dekken  but they are monetizing black theme now  :finnadie:

@buunguyen  Is it makes sense to do enterprise access also not free? I use this extension for a 5 year as long as it exists. 

As i understand this will not offence GPL licence if I fork this and release a version that allows darkmode and enterprise for free :godmode:? if some one already does this please give a link to the fork.

also, new sidebar hover over is  annoying ⚡️, u cannot prevent it to stay hidden, if u pin it and use hotkey it will turn on hover listener again

hanzheteng:
[...]

PR #649, opened 2019-04-14
feat: add hover-to-show option as default

nphi:
Address https://github.com/ovity/octotree/issues/648 and https://github.com/ovity/octotree/issues/651

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -86,7 +86,7 @@ $(document).ready(() => {
       let reload = false;
 
       Object.keys(changes).forEach((storeKey) => {
-        const [oldKeys, newKeys] = changes[storeKey];
+        const [oldValue, newValue] = changes[storeKey];
 
         switch (storeKey) {
           case STORE.TOKEN:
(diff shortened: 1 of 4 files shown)


buunguyen:
I think this should be an option: "Open sidebar on hover" (selected by default).

buunguyen:
Also, should make this the first option in the list.

nphi:
@buunguyen  I have updated the PR according to the requirement.

src/main.js
```diff
@@ -250,14 +252,32 @@ $(document).ready(() => {
            */
           isMouseInSidebar = true;
           clearTimer();
-          if (!isSidebarVisible()) toggleSidebar(true);
```
buunguyen: why remove this?

nphi: It causes unexpectedly opening the sidebar when hovering around the edge of the browser ( outside the toggler ).

buunguyen: But what's the effect of not having it? Why did we have it in the first place?

nphi: Screen Shot 2019-04-14 at 4 43 14 PM

There 2 explanations. 
 1. It has been there all the time and we've missed the edge case.
 2. The resizable border gets exposed ( check the screenshot ) and it causes the effect.

buunguyen: I'm afraid removing this will cause a minor annoyance. Currently, when the tree starts hiding, if I hover, it will expand again. But with this change, that won't happen correct? The tree will just collapse complete, then users would have to either hover or click the toggler.

nphi:
@buunguyen  I have addressed all your concerns.

Issue #654, opened 2019-04-14
Auto-fill password keeps popping up on Brave Browser


azdanov:
### Description
On every GitHub page a password auto-fill prompt gets displayed. It's probably related to API token field in octotree settings.

### Environment (if bug)

* Octotree version: 3.0.1
* Browser & version: Brave 0.62.51 Chromium: 73.0.3683.103 (Official Build) (64-bit)
* OS & version: macOS Mojave 10.14.4
* Screenshot:
<img width="520" alt="Screen Shot 2019-04-14 at 21 37 38" src="https://user-images.githubusercontent.com/6123841/56097516-fc0da400-5efd-11e9-9537-c58a2e038e1a.png">

__EDIT:__

Something to do with this:

```html
<input type="password" class="form-control input-block" data-store="TOKEN">
```

__EDIT2:__

By adding `autocomplete="off"` popup shouldn't get triggered. Needs some testing.

nphi:
@azdanov, I have tried your approach. Unfortunately, it doesn't work as expected. I will investigate more. Thanks for your report.

azdanov:
Would it be an acceptable solution to remove `type="password"`?

It is merely a visual deterrent if a malicious script wants to steal the token, it will do so regardless.

buunguyen:
@azdanov I agree. That should be fine. I made the update in 3.0.4. Hopefully it fixes the issue.

PR #656, opened 2019-04-15
fix: disable autocomplete in token input

nphi:
Address https://github.com/ovity/octotree/issues/654


nphi:
Still need more tests

Issue #659, opened 2019-04-15
Disable "Open sidebar on hover" doesnt work


ro0NL:
Version
3.0.1

Chrome  73.0.3683.103 

I've disabled the option "Open sidebar on hover", yet the sidebar stil seems to open when i hover above/below the "Octotree handle"

![image](https://user-images.githubusercontent.com/1047696/56115302-6147b080-5f63-11e9-9cea-4ee21fd9318f.png)

andywillis:
As an aside, I really like this tool, but, fwiw, the old UI for opening/closing the tool was much more useful and intuitive. "Click to open" is better than "click/hover to open, then pin to keep open", and "click to close" is better than "unpin and to wait for the sidebar to close".

nphi:
Hi, @ro0NL !

Somehow I couldn't reproduce the bug with the normal environment. It seems I'm missing something. Do you use other plugins ? It would be helpful if you can take a screenshot with console dev inspecting the DOM tree of the Octotree handle.

Thanks for reporting the issue.

YongYuH:
I use `Octotree` along with `ZenHub for GitHub` at the same time and encounter the same issue.

ro0NL:
@nphi i can confirm it's caused by a browser extension; in my case "[A Pixel Ruler For Web Developers](https://chrome.google.com/webstore/detail/designers-ruler/mlbnpnlmfngmlcmkhjpbfokdphfehhjj)" :joy:

nphi:
I can reproduce it now. :) I will make a PR to fix the issue ASAP.

buunguyen:
This should have been fixed in the latest release. If it still occurs, please let us know. Thanks for reporting the issue!

PR #662, opened 2019-04-15
fix: hide the resizable sidebar completely

nphi:
Address https://github.com/ovity/octotree/issues/659

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -23,6 +23,7 @@
 
   .ui-resizable-e {
     right: -4px;
+    width: 0px;
   }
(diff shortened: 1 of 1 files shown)


src/adapters/github.less
```diff
@@ -23,6 +23,7 @@
 
   .ui-resizable-e {
     right: -4px;
+    width: 0px;
```
buunguyen: Would this make it extremely difficult to resize? I mean previously width is 8px, there's a lot of space for users to hover and drag. Not sure if I miss anything.

Issue #650, opened 2019-04-14
Icon repeat


BigFaceMaster:
### Description
If this is a bug, provide as much information as you can to help reproduce the issue.

### Environment (if bug)

* Octotree version: v3
* Browser & version: Chrome 73.0.3683.103
* OS & version: MacOS Mojave 10.14.4
* Screenshot, if any (drag an image here) 
![image](https://user-images.githubusercontent.com/15710933/56087315-1a8e8380-5e9b-11e9-87de-95a3aa50ce2f.png)

* Console error log below this line

Seasonley:
It conflicts with `Github File Icons` extensions

nphi:
@BigFaceMaster I am unable to reproduce the issue. Could you investigate more details and describe steps to reproduce it so that we could fix the issue ASAP?

Thanks for your cooperation.

![Screen Shot 2019-04-14 at 3 13 59 PM](https://user-images.githubusercontent.com/8633840/56090159-48d68800-5ec8-11e9-9163-6d7a0c282a73.png)

BigFaceMaster:
@Seasonley thank you guys, 
@nphi  as @Seasonley say: `It conflicts with Github File Icons extensions`, and I removed the `Github File Icons` and use ''File Icon for GitHub, GitLab and Bitbucket"

duylam:
Sent a PR to https://github.com/lvarayut/github-file-icons/pull/4

buunguyen:
PR is merged in github-file-icons. Please download the new version of it to have this issue fixed.

PR #670, opened 2019-04-15
Fix: file icon is renderred incorrectly

duylam:
### Problem

Fix: https://github.com/ovity/octotree/issues/650

### Screenshots

Screen Shot 2019-04-15 at 11 45 23 PM


buunguyen:
Looks like something to be fixed in the File Icons extension instead.

buunguyen:
Please notify the owner of that repo.

PR #679, opened 2019-04-17
enhance: add escape to close and vimium compatibility

nphi:


src/main.js
```diff
@@ -232,7 +234,7 @@ $(document).ready(() => {
       let timerId = null;
 
       const startTimer = (delay) => {
-        if (!isMouseInSidebar && !isSidebarPinned()) {
+        if (this.store.get(STORE.HOVEROPEN) && !isMouseInSidebar && !isSidebarPinned()) {
```
buunguyen: Not sure about this check. If not hover-open, meaning users might not know how to close. I think we should still close automatically in float mode.

src/main.js
```diff
@@ -215,8 +215,10 @@ $(document).ready(() => {
       handleHoverOpenOption(this.store.get(STORE.HOVEROPEN));
 
       // Immediately closes if click outside the sidebar.
-      $document.on('click', () => {
-        if (!isMouseInSidebar && !isSidebarPinned() && isSidebarVisible()) {
+      $document.on('click keyup', (e) => {
+        const isEscape = e.type === 'keyup' && e.key === 'Escape';
```
buunguyen: Did someone request this? Not sure the implication... The more complicated this code is, the more nervous I am.

nphi: It's just a suggestion.

src/main.js
nphi: If user prefer to clicking instead of hovering, I think clicking to close would make more sense. Users will intuitively know it because the behaviour is similar to other apps.

buunguyen: I agree. The only probably is we hide the toggle, so at least initially users don't know they need to click in the page to close. Not sure if it could confuse first time users and cause uninstallation.

nphi: for initial users, they use hovering as default so the timer still works.

buunguyen: I still think it's confusing that the behavior is very implicit. Let's not implement it for now until we're clear this is desirable.

PR #692, opened 2019-04-21
enhancement: prevent full local storage issue

nphi:
- Add a error message to notify user when the local storage is full
- Prevent the potential of full local storage when caching hug repo


buunguyen:
Don't like this approach, too verbose and brittle. Just try/catch the storage set, then call handleError with specific flag for local storage, don't "abuse" the 206.

PR #699, opened 2019-04-24
fix: show error message when full local storage

nphi:

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -197,7 +197,11 @@ class Adapter {
         }
 
         break;
+      case 600:
+        error = 'Full Local Storage';
+        message = 'The local storage is full. You need to clear it and retry.';
 
+        break;
       // Fallback message
       default:
         error = message = jqXHR.statusText;
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -39,9 +39,10 @@ const GH_404_SEL = '#parallax_wrapper';
 // When Github page loads at repo path e.g. https://github.com/jquery/jquery, the HTML tree has
 // <main id="js-repo-pjax-container"> to contain server-rendered HTML in response of pjax.
 // However, that <main> element doesn't have "id" attribute if the Github page loads at specific
-// file e.g. https://github.com/jquery/jquery/blob/master/.editorconfig.
+// File e.g. https://github.com/jquery/jquery/blob/master/.editorconfig.
 // Therefore, the below selector uses many path but only points to the same <main> element
-const GH_PJAX_CONTAINER_SEL = '#js-repo-pjax-container, div[itemtype="http://schema.org/SoftwareSourceCode"] main, [data-pjax-container]';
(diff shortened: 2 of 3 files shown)


src/adapters/github.js
```diff
@@ -342,16 +347,22 @@ class GitHub extends PjaxAdapter {
...
-              const oldestRepo = repos.reduce((min, p) => (hugeRepos[p] < hugeRepos[min] ? p : min));
-              delete hugeRepos[oldestRepo];
-            }
-            hugeRepos[repo] = new Date().getTime();
-            this.store.set(STORE.HUGE_REPOS, hugeRepos);
+
+          const handleHugeRepo = (isFullStorage) => this._handleError(cfg, {status: isFullStorage ? 600 : 206}, cb);
+
+          // The hugeRepos isn't a default {} due to the full Local Storage
+          if (!hugeRepos) return handleHugeRepo(true);
```
buunguyen: Why should store.get() return wrong data when storage is full? Are you sure about this?

src/adapters/github.js
```diff
@@ -342,16 +347,22 @@ class GitHub extends PjaxAdapter {
...
-          if (!hugeRepos[repo]) {
-            // If there are too many repos memoized, delete the oldest one
-            if (repos.length >= GH_MAX_HUGE_REPOS_SIZE) {
-              const oldestRepo = repos.reduce((min, p) => (hugeRepos[p] < hugeRepos[min] ? p : min));
-              delete hugeRepos[oldestRepo];
-            }
-            hugeRepos[repo] = new Date().getTime();
-            this.store.set(STORE.HUGE_REPOS, hugeRepos);
+
+          const handleHugeRepo = (isFullStorage) => this._handleError(cfg, {status: isFullStorage ? 600 : 206}, cb);
```
buunguyen: Why 'handleHugeRepo' calls handleError? Just doesn't make any sense.

src/adapters/adapter.js
```diff
@@ -197,7 +197,11 @@ class Adapter {
         }
 
         break;
+      case 600:
```
buunguyen: Isn't 600 still valid HTTP code? Should be something that outside any range.

src/adapters/github.js
nphi: When the new user installs or update the Octotree v3 but his local storage is already full, the default setting is unable to set storage, which leads to the variable is undefined

buunguyen:
Actually if there's no bug in storage code. I don't think we should do too much. We already log the error.

src/adapters/github.js
buunguyen: In this case, should we just short circuit the code to return early, skipping the check huge repo logic? That would make things simple without a big and ugly change like this (now if it's less ugly, I'm okay, but this is really ugly).

src/adapters/github.js
nphi: I will think a better name. Basically, the function just determines whether it should pass the full local storage status or the huge repo status to the `handleError`

PR #703, opened 2019-04-26
Show full icon branch

Hongarc:
Before:
image

After:
image

This commit will show full icon branch.

I am using lastest chrome version.

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -158,6 +158,7 @@
         .octotree-header-branch {
           line-height: 1;
           margin-left: -1px;
+          padding-bottom: 1px;
           font-size: 11px;
           color: white;
 


nphi:
That's a good catch. But I'd prefer to changing the top of `.octotree-icon-branch` which causes the hidden. Could you change the solution a little bit ?

Thanks for your effort and contribution.

buunguyen:
I made the change myself as I need to release an update with other bugs today.

@Hongarc thanks for finding the issue, super good catch!

Hongarc:
> That's a good catch. But I'd prefer to changing the top of `.octotree-icon-branch` which causes the hidden. Could you change the solution a little bit?

I tested in `.octotree-icon-branch` but it didn't work.
So I change in `.octotree-header-branch`.

Hongarc:
@nphi @buunguyen 
I think to remove `overflow: hidden;` in `.octotree-header-branch`	is better idea.

Hongarc/octotree@ccd37ab

PR #705, opened 2019-04-27
Show full icon branch

Hongarc:
Same #703 
I thinked da2e6c0 fixed it.

But when I zoom my chrome to 110% and 90%, I saw that issue again(that commit work for 120% and 80%)

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -158,6 +158,7 @@
         .octotree-header-branch {
           line-height: 1;
           margin-left: -1px;
+          padding-bottom: 1px;
           font-size: 11px;
           color: white;
 


src/adapters/github.less
```diff
@@ -124,13 +124,13 @@
         .octotree-header-repo,
         .octotree-header-branch {
           width: 100%;
-          overflow: hidden;
```
buunguyen: Why delete this? Thought this is needed in case branch name is long.

Hongarc: Oh, my fault.

I think we should change `top: 0px` in `.octotree-icon-branch`(or `padding-bottom: 1px;` `.octotree-header-branch`)

Hongarc:
@buunguyen You change `top: 1px` so we need to add `padding-bottom` to  1px

buunguyen:
For some reason this doesn't fix the problem. I still see the cut-off when zooming.

PR #708, opened 2019-04-27
fix: pinned sidebar overlay the content in small screen

nphi:

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -48,6 +48,8 @@ const GH_CONTAINERS = '.container, .container-lg, .container-responsive';
 const GH_HEADER = '.js-header-wrapper > header';
 const GH_RAW_CONTENT = 'body > pre';
 const GH_MAX_HUGE_REPOS_SIZE = 50;
+const GH_HIDDEN_RESPONSIVE_CLASS = '.d-none';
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -107,7 +107,13 @@ class GitHub extends PjaxAdapter {
...
     const $header = $(GH_HEADER);
     const $containers = $(GH_CONTAINERS);
-    const autoMarginLeft = ($(document).width() - $containers.width()) / 2;
+
+    /**
+     * The header is hidden in the small screen (responsive mode) while the footer is always displayed.
+     * Using $containers.last().width() (footer) instead of $containers.width() (header)
+     * to get correct container width.
+     */
+    const autoMarginLeft = ($(document).width() - $containers.last().width()) / 2;
```
buunguyen: $container is $(GH_CONTAINERS). If you do this, should you change the value of GH_CONTAINERS instead?

buunguyen: Also, how to ensure the code will still work if DOM change? 
And the lines below that use $containers, do they still work? Do you need to add .last()? Why not?

Basically, I'm not comfortable with this change. Need to make sure the code is robust instead of making it feel like hack for one particular issue.

PR #710, opened 2019-04-27
Rename HISTORY to CHANGELOG

Hongarc:
I think `CHANGELOG` is better.


buunguyen:
Thanks, but we link to HISTORY from a few places so need to do more than just renaming the file. Not sure it's worth it though.

Issue #729, opened 2019-05-24
Will the Cache be available soon?


gixid192:
I create this with the hope to know the whether you guys will implement the cache feature.
This is not a new issue, it has been around since 2014 (#104) and several related issues (#442, #489...)

With the new Cache API, I implemented my own version in my machine, seems to work well.
Please let me know the status of this feature, I will try to make a PR if needed.

buunguyen:
No, we have discussed internally several times but haven't really got the time to commit to it. The biggest challenge is, unsurprisingly, when to invalidate the cache, e.g. the user adds a file, another user deletes a file etc. Maybe it's an option with a warning like "You won't see changes by other users in real time" when enabling cache. Any idea is welcome and a PR will of course be totally awesome.

gixid192:
I personally don't work in the case a repo needed to update that frequently.
I also don't work largely on Github so could you confirm if the case you mentioned is a normal case?

I originally wanted to create a PR but then because of #442, I don't think it's a PR, it's just my personal code, not sure if it fit for everyone.
My current idea is attach a hash `#frc=1` in the URL so that the cache will be forced to reload if needed.
The cache will also automatically reloaded if it is `old enough`, here I hard-coded this to 30 days period. This should be the option for users and even more robust, for each of the repos.

What do you think about my approach? Should I create a PR anyway? Or just share the code via pastebin?

buunguyen:
Maybe a good compromise is providing an option, "Cache repositories for 7 days". When caching is enabled, there should be a way for users to force cache invalidation. Maybe add an icon in the header (before to settings) to reload repo if caching is enabled?

A PR would be nice.

gixid192:
I just created a PR.
To be honest, the code need to much improve, I just want to demonstrate my idea.

buunguyen:
@gixid192 	thanks for creating the PR! We'll review and let you know any feedback.

PR #731, opened 2019-05-24
Add Cache API

gixid192:
### Problem
Attemping to resolve the cache issue (#729)

### Solution
Added the use of new Cache API

Edit:
### Side note
The code is where I try to demonstrate the idea, it is not a perfect code.

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -5,6 +5,24 @@ class Adapter {
     this.store = store;
   }
 
+  _setCacheTime(storeName, cacheKey){
+    if(!this.store.get(storeName)){
+        this.store.set(storeName, {});
+    }
+    const cachedRepos = this.store.get(storeName);
+    cachedRepos[cacheKey] = new Date().getTime();
+    this.store.set(storeName, cachedRepos);
+  }
+
+  _reloadCache(storeName, cacheKey, cb, reloadAfterDay = 29){
+    const cachedRepos = this.store.get(storeName);
+    const cachedRepoTime = cachedRepos[cacheKey];
+    const timeDiff = (cachedRepoTime - new Date().getTime()) / (1000 * 60 * 60 * 24);
+    if(timeDiff > reloadAfterDay){
+      cb()
+    }
+  }
+
   /**
    * Loads the code tree of a repository.
    * @param {Object} opts: {
@@ -24,6 +42,17 @@ class Adapter {
 
     opts.encodedBranch = opts.encodedBranch || encodeURIComponent(decodeURIComponent(repo.branch));
 
+    const host =
+      location.protocol + '//' + (location.host === 'github.com' ? 'api.github.com' : location.host + '/api/v3');
+    const CACHE_KEY = `${host}/repos/${opts.repo.username}/${opts.repo.reponame}${path || ''}`;
+    const CACHE_NAME = 'loaded-tree';
+    const STORE_NAME = 'loaded-tree-storage';
+    let force;
+    if(location.hash.search('frc=') !== -1){
+        force = true;
+    }
+    let globalCache;
+
     this._getTree(path, opts, (err, tree) => {
       if (err) return cb(err);
 
@@ -40,6 +69,9 @@ class Adapter {
 
             // We're done
             if (item === undefined) {
+              this._setCacheTime(STORE_NAME, CACHE_KEY);
+              globalCache.put(CACHE_KEY, new Response(JSON.stringify(folders[''])));
+
               return cb(null, folders['']);
             }
 
@@ -144,7 +176,19 @@ class Adapter {
           setTimeout(() => nextChunk(iteration + 1));
         };
 
-        nextChunk();
+        caches.open(CACHE_NAME).then(cache => {
+          globalCache = cache
+          cache.match(CACHE_KEY).then(currentMatch => {
+            if(typeof currentMatch === "undefined" || force){
+                nextChunk();
+            } else {
+              currentMatch.text().then(function(data){
+                return cb(null, JSON.parse(data))
+              });
+              this._reloadCache(STORE_NAME, CACHE_KEY, nextChunk);
+            }
+          })
+        });
       });
     });
   }
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -340,7 +340,7 @@ class GitHub extends PjaxAdapter {
     });
   }
 
-  _get(path, opts, cb) {
+  _get(path, opts, cb, force) {
     const host =
       location.protocol + '//' + (location.host === 'github.com' ? 'api.github.com' : location.host + '/api/v3');
     const url = `${host}/repos/${opts.repo.username}/${opts.repo.reponame}${path || ''}`;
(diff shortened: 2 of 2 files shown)


src/adapters/github.js
```diff
@@ -350,24 +350,51 @@ class GitHub extends PjaxAdapter {
...
-        } else cb(null, data);
+    const CACHE_NAME = 'cached-repo-v1';
+    const STORE_NAME = 'cached-repo-time';
+    const CACHE_KEY = url;
+    if(location.hash.search('frc=') !== -1){
+        force = true;
+    }
+    caches.open(CACHE_NAME).then(cache => {
+      cache.match(CACHE_KEY).then(currentMatch => {
+        if(typeof currentMatch === "undefined" || force){
```
crashbell: Please follow the coding convention with spaces.

`if (`, `) {`.

src/adapters/adapter.js
```diff
@@ -5,6 +5,24 @@ class Adapter {
...
+    }
+    const cachedRepos = this.store.get(storeName);
+    cachedRepos[cacheKey] = new Date().getTime();
+    this.store.set(storeName, cachedRepos);
+  }
+
+  _reloadCache(storeName, cacheKey, cb, reloadAfterDay = 29){
+    const cachedRepos = this.store.get(storeName);
+    const cachedRepoTime = cachedRepos[cacheKey];
+    const timeDiff = (cachedRepoTime - new Date().getTime()) / (1000 * 60 * 60 * 24);
```
crashbell: Looks like a magic number here, I think we should constant.

It's not clear to me. Invalidating the cache after 24 hours seems not a good approach. Do we have any other ways to invalidate the cache? The code in repo may change during that period, how can we manage that change?

gixid192: To be clear: I did not set it to 24 hours. It's the `reloadAfterDay` value: 30 days.

What you concern is already addressed by Buu in his comment: https://github.com/ovity/octotree/issues/729#issuecomment-495483665

What do you think about this approach? To me, I really like his idea.

src/adapters/adapter.js
```diff
@@ -24,6 +42,17 @@ class Adapter {
 
     opts.encodedBranch = opts.encodedBranch || encodeURIComponent(decodeURIComponent(repo.branch));
 
+    const host =
+      location.protocol + '//' + (location.host === 'github.com' ? 'api.github.com' : location.host + '/api/v3');
```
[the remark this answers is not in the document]
gixid192: Totally Agree!
Please forgive me, I didn't check the whole project. I just try to demonstrate the idea through the code.

This should not be merged as a good PR (I already said it in the issue)

src/adapters/github.js
gixid192: I read the contribution section in the README, however, not clear how to follow. Do you guys use Prettier or something? How do I know what style that the code is using? Or is it just obvious by looking at the code?
Sorry for the silly questions, this is my first PR, I'm not usually contribute to projects on Github.

src/adapters/adapter.js
```diff
@@ -40,6 +69,9 @@ class Adapter {
 
             // We're done
             if (item === undefined) {
+              this._setCacheTime(STORE_NAME, CACHE_KEY);
+              globalCache.put(CACHE_KEY, new Response(JSON.stringify(folders[''])));
```
buunguyen: When the "Load entire tree at once" is unchecked (`opts.node` is undefined) , the cache shouldn't be used. Maybe should handle that case?

src/adapters/adapter.js
```diff
@@ -144,7 +176,19 @@ class Adapter {
           setTimeout(() => nextChunk(iteration + 1));
         };
 
-        nextChunk();
+        caches.open(CACHE_NAME).then(cache => {
+          globalCache = cache
+          cache.match(CACHE_KEY).then(currentMatch => {
+            if(typeof currentMatch === "undefined" || force){
```
buunguyen: Maybe we should check `force` before going into this. That is:

```
if (force || !opts.node /* as commented above, skip cache code if lazy load tree */) nextChunk();
else {
  caches.open()...
}

File: src/adapters/adapter.js [hunk 6]
```diff
@@ -24,6 +42,17 @@ class Adapter {
 
     opts.encodedBranch = opts.encodedBranch || encodeURIComponent(decodeURIComponent(repo.branch));
 
+    const host =
+      location.protocol + '//' + (location.host === 'github.com' ? 'api.github.com' : location.host + '/api/v3');
+    const CACHE_KEY = `${host}/repos/${opts.repo.username}/${opts.repo.reponame}${path || ''}`;
+    const CACHE_NAME = 'loaded-tree';
+    const STORE_NAME = 'loaded-tree-storage';
+    let force;
+    if(location.hash.search('frc=') !== -1){
```
buunguyen: This is nice but we will need something in the UI for users to trigger this. Maybe add an icon in place of the loading icon to do trigger refresh.

src/adapters/adapter.js
```diff
@@ -5,6 +5,24 @@ class Adapter {
...
+    cachedRepos[cacheKey] = new Date().getTime();
+    this.store.set(storeName, cachedRepos);
+  }
+
+  _reloadCache(storeName, cacheKey, cb, reloadAfterDay = 29){
+    const cachedRepos = this.store.get(storeName);
+    const cachedRepoTime = cachedRepos[cacheKey];
+    const timeDiff = (cachedRepoTime - new Date().getTime()) / (1000 * 60 * 60 * 24);
+    if(timeDiff > reloadAfterDay){
+      cb()
```
buunguyen: When force reload or expire, I think we should delete the cache, no? Otherwise, user cache will keep growing forever. 

Related is: if users stop accessing a repo (that was cached), do browsers clean up cache automatically? Or the cache will keep growing...

src/adapters/adapter.js
buunguyen: Also, maybe we should provide an option to users, like "Cache frequently accessed repos"? So they can opt-in only if they want to speed up the repos they usually work with (e.g. the last 10 repos that are accessed more than 10 times).

buunguyen:
I need to rethink this feature. The current approach seems very complicated with many edge cases not handled, as pointed out in the comments (and they are probably not exhaustive).

src/adapters/adapter.js
gixid192: Agree, UI is much better.

src/adapters/adapter.js
gixid192: I didn't consider the case where users stop using the repo. 
I borrowed this cache idea from PWA, from my experience, the service worker will handle the cache.

In this case, to be honest, I don't know how the browsers will handle it. Reading from [this](https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/offline-for-pwa#how_can_i_tell_how_much_storage_space_my_app_is_using) article, `Cache API falls into the “best effort” category which means the storage can be cleared without interrupting the user`, they don't provide more information about how it work.

If the Cache API is still not being used widely, a workaround for this maybe is to let the users know what repos that they're having the cache version and how many days have the cache been saved.

src/adapters/adapter.js
gixid192: thanks! my bad.

src/adapters/adapter.js
gixid192: I haven't checked the case with "Load entire tree at once" 😅

You know it better than me so I'm all ears, providing users with more options to control is good UX too.

PR #739, opened 2019-07-03
Refactor: add tree item interceptor

nphi:

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -74,20 +74,7 @@ class Adapter {
             }
 
             if (item.patch) {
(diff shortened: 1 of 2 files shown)


src/adapters/adapter.js
```diff
@@ -84,6 +84,9 @@ class Adapter {
                 : '';
               patch += additions !== 0 ? `<span class="text-green">+${additions}</span>` : '';
               patch += deletions !== 0 ? `<span class="text-red">-${deletions}</span>` : '';
+
+              patch = this.editPatch(path, item);
```
buunguyen: Let's move all logic from 77-88 into this.buildPatchHtml()

PR #745, opened 2019-07-21
Update css jstree anchor and octotree patch

nphi:

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -179,6 +179,8 @@
         outline: none;
         padding-right: 0;
         text-decoration: none;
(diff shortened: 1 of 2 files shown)


src/adapters/github.less
```diff
@@ -179,6 +179,8 @@
         outline: none;
         padding-right: 0;
         text-decoration: none;
+        width: calc(100% - 20px);
```
buunguyen: I suppose file name will be cut off. Should we show ...?

nphi: @buunguyen  Done a. It's good to go.

PR #749, opened 2019-07-25
fix issue incorrect sort version number

ngocpt1906:
Problem
The sort function is incorrect if Folder/File name contain version number

Solution
Add cmpVersion that support compare version number

Screenshots
61864503-bf3d4a00-aefb-11e9-91ad-8c322f709b5b

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -5,15 +5,15 @@ class TreeView {
     this.$view = $dom.find('.octotree-tree-view');
     this.$tree = this.$view
       .find('.octotree-view-body')
-      .on('click.jstree', '.jstree-open>a', ({target}) => {
+      .on('click.jstree', '.jstree-open>a', ({ target }) => {
         setTimeout(() => this.$jstree.close_node(target));
       })
-      .on('click.jstree', '.jstree-closed>a', ({target}) => {
+      .on('click.jstree', '.jstree-closed>a', ({ target }) => {
         setTimeout(() => this.$jstree.open_node(target));
       })
       .on('click', this._onItemClick.bind(this))
       .jstree({
-        core: {multiple: false, animation: 50, worker: false, themes: {responsive: false}},
+        core: { multiple: false, animation: 50, worker: false, themes: { responsive: false } },
         plugins: ['wholerow', 'search']
       });
   }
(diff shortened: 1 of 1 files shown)


src/view.tree.js
```diff
@@ -5,15 +5,15 @@ class TreeView {
     this.$view = $dom.find('.octotree-tree-view');
     this.$tree = this.$view
       .find('.octotree-view-body')
-      .on('click.jstree', '.jstree-open>a', ({target}) => {
+      .on('click.jstree', '.jstree-open>a', ({ target }) => {
```
danhnguyen30041975: Need to revert such of changes

src/view.tree.js
```diff
@@ -93,9 +93,25 @@ class TreeView {
     return str && str.replace(/[<>'"&]/g, '-');
   }
 
+  cmpVersions(a, b) {
+    let i, diff;
+    let regExStrip0 = /(\.0+)+$/;
+    let segmentsA = a.replace(regExStrip0, '').split('.');
```
danhnguyen30041975: Need to address sorting in abstract way

src/view.tree.js
```diff
@@ -93,9 +93,25 @@ class TreeView {
     return str && str.replace(/[<>'"&]/g, '-');
   }
 
+  cmpVersions(a, b) {
+    let i, diff;
+    let regExStrip0 = /(\.0+)+$/;
```
buunguyen: We use `const` for these

src/view.tree.js
```diff
@@ -93,9 +93,25 @@ class TreeView {
     return str && str.replace(/[<>'"&]/g, '-');
   }
 
+  cmpVersions(a, b) {
+    let i, diff;
```
buunguyen: Don't declare loop variable outside. Please read existing code so that you follow the same style instead trying a new style.

src/view.tree.js
```diff
@@ -93,9 +93,25 @@ class TreeView {
     return str && str.replace(/[<>'"&]/g, '-');
   }
 
+  cmpVersions(a, b) {
```
buunguyen: Why is this named "compareVersion"? This is supposed to compare files. Version is just a part of it.

PR #751, opened 2019-07-28
Refactor: pass global utils to plugin

nphi:


buunguyen:
Don't do this. Refactor deXss.js to something like:

```
let $dummyDiv;
window.deXss = (str) => {
  $dummyDiv =   $dummyDiv || $('<div></div>');
  return this._$dummyDiv.text(str).html();
}
```

Issue #543, opened 2018-09-25
Feature request: Enhance options where to show


basejumpa:
Feature request: Enhance options where to show.

My usecase is to show it in the context of pull-requests only. So suggest to replace the option "Snow in non-code pages" by a field of on/off options:

```
Show in:
* ( ) Code
* ( ) Code/Commits
* ( ) Pull request
```

danhnguyen30041975:
IMO, below options make sense

```
Show in:
* [ ] Code
* [ ] Pull Request
* [ ] Commit
* [ ] Other pages
```

Note: `[ ]` is checkbox

buunguyen:
Change to `Show sidebar in:`

Should select all options by defaults. Also move this section to after Shortkey.

buunguyen:
This is in 3.0.10.

PR #760, opened 2019-08-11
Enhance options where to show

endyquang:
### Problem
Implement options where to show as stated in #543 

### Solution
Change to
```
Show in:
 [ ] Code
 [ ] Pull Request
 [ ] Commit
 [ ] Other pages
```
The problem is when no option is chosen, the octotre would dispear so i chose to make octotree always appears on the default code page of repo

### Screenshots
image


buunguyen:
Thanks for this. I think the UI can be improved a bit. Some suggestions:

* "Show in" => "Show only in pages"
* "Other pages" => "Others"
* Can we try to show the options in just a single column instead of 2? I think 2 columns might have problem if users resize the sidebar to min width. Also, if there's a way to make it clear this is the same section (different from remaining checkboxes) it would be nice. Maybe indent the show-in checkboxes and add a left border to the group? Or have a different background? Other idea is welcome.

Also, by default, should check all options.

endyquang:
Oops i didnt notice that it would break if users resize.
Then i guess it should be displayed in another way, like you suggested. What do you think about this one?

image

Btw all options should be checked, if not try remove the existing keys in localStorage

src/main.js
```diff
@@ -88,23 +88,26 @@ $(document).ready(() => {
...
-            reload = true;
-            break;
-          case STORE.HOVEROPEN:
-            handleHoverOpenOption(newValue);
-            break;
-          case STORE.HOTKEYS:
-            setHotkeys(newValue, oldValue);
-            break;
+      const reactions = {
+        [STORE.TOKEN] () {},
```
buunguyen: This mighe be a bug. Token / LoadAll / Icons / Pull Requests should all trigger reload. But look like with this logic, it doesn't do that.

src/adapters/github.js
```diff
@@ -159,12 +159,19 @@ class GitHub extends PjaxAdapter {
       return cb();
     }
 
-    // Check if we should show in non-code pages
-    const isPR = type === 'pull';
-    const isCodePage = !type || isPR || ['tree', 'blob', 'commit'].indexOf(type) >= 0;
-    const showInNonCodePage = this.store.get(STORE.NONCODE);
-    if (!showInNonCodePage && !isCodePage) {
-      return cb();
+    if (type) {
```
buunguyen: Wonder if this preserves current logic. If `type` doesn't exist, should false into the case of "Others", in which case if "Others" is checked, the tree should show, correct?

endyquang: in what i understand, the "type" variable would be null/undefined if it's on the default repo page. I want to make it always visible on this page regard of the setting, thus i add this condition

src/main.js
endyquang: Oh yes it is, do you think i should remove the lines 92, 93, 94?

src/adapters/github.js
buunguyen: Default page would be `code`. So maybe change the hash code to empty string so that it matches type? Basically I want code to be more explicit. Currently too implicit, will have to stop and think about the flow.

endyquang: Default page is `code`? Does this means it would be shown on the `code page` whether the `show in code` option is check, or shown when all options are unchecked? And i admit that this is too implicit

src/main.js
endyquang: or just use the old switch case logic would be easier to read

buunguyen: Agreed

src/adapters/github.js
buunguyen: Yes, it should show ONLY when show in code is checked. The idea is some people might not even want to see the tree in code page, only in pull request page.

buunguyen:
Example of "code" page (i.e. Code tab): 

* https://github.com/ovity/octotree 
* https://github.com/ovity/octotree/tree/master

Example of "commit" page: https://github.com/ovity/octotree/commit/ec51416bc4c1fba87c1bbe620e61e1f7a2b3481d

Examples of "pull request" page: 

* https://github.com/ovity/octotree/pull/760/<BLANK OR ANYTHING>  

Rest is "others"

@duylam do I miss any case? I think I do but can't think too clearly right now.

src/adapters/github.js
endyquang: ONLY when show in code is checked? vs Default page is `code`?
I dont get it. I'm thinking of the situation when no options are checked, the octotree disappear & can't be turn back on

buunguyen: That's a good point... Hmm.. On the other hand, if people don't want to see it in code page and it still shows, would be quite weird. How about we force at least 1 option is selected? Other idea is welcome.

buunguyen: Can you pick another issue to work on while I think about this?

endyquang: Sure, let me know when you think through

duylam: > How about we force at least 1 option is selected? Other idea is welcome.

That's one option, I agree. And we need a little validation hooks on checkbox

And I have another option, it can save us from writing validation but looks a bit odd

```
Show in:
( ) Every page
( ) Only at Code / Commit page
( ) Only at Pull Request page
( ) Only at Code / Commit and Pull Requests pages
```

`( )` is ratio button. For now, I think we can go with the ratio-based approach until we have more visibility options

buunguyen: Great idea @duylam.

@endyquang let's do this:

```
Enable Octotree in:
 ( ) All pages
 ( ) Code/commits & pull requests
 ( ) Code/commits
 ( ) Pull requests
``` 

Rules:

* All pages is default
* Code / commit: /ovity/octotree/<BLANK or tree/xxx or blob/xxx or commit/xxx> (but not /commits)
* Pull request: /ovity/octotree/pull/xxx (but not /pulls)

src/main.js
endyquang: Yep agreed

src/adapters/github.js
endyquang: @duylam @buunguyen What would happen if none of the above options are checked?

buunguyen: They're radio options not checkboxes, so one must be checked.

buunguyen: Logic should be fail-safe, i.e. even if nothing in storage, set the All pages to true.

endyquang: I get it, but with these radios, it would not be as flexible as with 4 checkboxes. I have updated and made a validation hook, let me commit it first

buunguyen: Why? These radios cover ALL options if you read carefully.

endyquang: I guess, in case users want `code only` or `commit only` or `commit & PR` or `code & PR`. Are there such cases?

buunguyen: I see where you come from. The idea is code / commit are treated the same as it's part of Code tab. Basically simplify the life of users a bit, at the expense of a little bit less flexible. But I think it's not a bad trade off.

endyquang: Okay i believe now. Let me update everthing & push another commit

endyquang: @buunguyen Done, please check again

buunguyen: Thanks, can you screenshot? Also, please help test carefully.

src/adapters/github.js
```diff
@@ -159,12 +159,20 @@ class GitHub extends PjaxAdapter {
...
-      return cb();
+
+    // Skip rendering the octotree in the unselected pages
+    if (!this.store.get(STORE.SHOWIN_ALL)) {
+      const isCodeCommit = !type || ['tree', 'blob', 'commit'].includes(type);
+      const isCodeCommitPR = isPR || isCodeCommit;
+
+      if (
+        (this.store.get(STORE.SHOWIN_CODE_COMMIT_PR) && !isCodeCommitPR) ||
+        (this.store.get(STORE.SHOWIN_CODE_COMMIT) && isCodeCommit) ||
```
buunguyen: Is this logic correct? Thought it should be `!isCodeCommit`, same with next line. If I'm wrong, then the entire thing is hard for me to understand, would need refactor.

src/adapters/github.js
endyquang: image
Yep i'll be carefull

src/adapters/github.js
endyquang: Oops sorry, thank you for pointing that out

PR #766, opened 2019-08-17
truncate filename if it is overflow

nphi:
Screen Shot 2019-08-17 at 6 24 47 PM


buunguyen:
Nice. But the first line: the directory name looks funny, why so many "..."s?

nphi:
> Nice. But the first line: the directory name looks funny, why so many "..."s?

buunguyen:
@nphi I think you can address that by updating the CSS selector so that it applies to the last sibling within a parent.

nphi:
a @buunguyen, it wouldn't work if other siblings contain long text. Truncating the last sibling is not enough.

PR #767, opened 2019-08-18
add branch switching

Fahme:
### Problem
support for branch switching in octotree
### Solution
This is a simple solution to support branch switching using octotree, not yet complete, UI not in it's best condition, but open to discuss more and try to bring this solution to a merge :).
### Screenshots

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -265,6 +265,15 @@ class Adapter {
   }
 
   /**
+   * Gets repo branches.
+   * @param {Object} opts - {opts}
+   * @api public
+   */
+  getBranch(opts) {
+    throw new Error('not implemnited');
+  }
+
+  /**
    * Selects the file at a specific path.
    * @api public
    */
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -269,8 +269,8 @@ class Adapter {
    * @param {Object} opts - {opts}
    * @api public
    */
-  getBranch(opts) {
-    throw new Error('not implemnited');
+  getBranches(opts) {
+    throw new Error('Not implemented');
   }
 
   /**
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -108,7 +108,7 @@ class GitHub extends PjaxAdapter {
   }
 
   // @override
-  getBranch(opts) {
+  getBranches(opts) {
     return new Promise((resolve, reject) => {
       this._get('/branches', opts, (err, res) => {
         if (err) reject(err);
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -108,6 +108,16 @@ class GitHub extends PjaxAdapter {
   }
 
   // @override
+  getBranch(opts) {
+    return new Promise((resolve, reject) => {
+      this._get('/branches', opts, (err, res) => {
+        if (err) reject(err);
+        else return resolve(res);
+      });
+    });
+  }
+
+  // @override
   updateLayout(sidebarPinned, sidebarVisible, sidebarWidth) {
     const SPACING = 10;
     const $header = $(GH_HEADER);
@@ -354,13 +364,11 @@ class GitHub extends PjaxAdapter {
 
     if (path && path.startsWith('http')) {
       url = path;
-    }
-    else {
+    } else {
       const host =
         location.protocol + '//' + (location.host === 'github.com' ? 'api.github.com' : location.host + '/api/v3');
       url = `${host}/repos/${opts.repo.username}/${opts.repo.reponame}${path || ''}`;
     }
-
     const cfg = {url, method: 'GET', cache: false};
 
     if (opts.token) {
--- a/src/main.js
+++ b/src/main.js
@@ -12,7 +12,7 @@ $(document).ready(() => {
     const $pinner = $sidebar.find('.octotree-pin');
     const adapter = new GitHub(store);
     const treeView = new TreeView($dom, store, adapter);
-    const branchView = new BranchSwitch($dom);
+    const branchView = new BranchSwitch($dom, $html);
     const optsView = new OptionsView($dom, store, adapter);
     const helpPopup = new HelpPopup($dom, store);
     const errorView = new ErrorView($dom, store);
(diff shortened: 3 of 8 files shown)


src/adapters/adapter.js
```diff
@@ -264,6 +264,15 @@ class Adapter {
     throw new Error('Not implemented');
   }
 
+  /**
+   * Gets repo branches.
+   * @param {Object} opts - {opts}
+   * @api public
+   */
+  getBranch(opts) {
+    throw new Error('not implemnited');
```
nphi: typo

src/main.js
```diff
@@ -117,9 +121,9 @@ $(document).ready(() => {
       }
     }
 
-    function tryLoadRepo(reload) {
+    function tryLoadRepo(reload, isChangedBranch = false, branch) {
       const token = octotree.getAccessToken();
-      adapter.getRepoFromPath(currRepo, token, (err, repo) => {
+      adapter.getRepoFromPath(currRepo, token, async (err, repo) => {
```
nphi: We don't need to use `async` if there is no `await` in the body

src/view.branch.js
```diff
@@ -0,0 +1,61 @@
...
+    if (!this.branches.length) {
+      this.branches = branches;
+    }
+  }
+
+  _showPopUp() {
+    $('html')
+      .append(
+        `
+           <div class='modal-root'>
```
nphi: could we mimic the dropdown of Github branches switching for the sake of consistency ?

src/view.branch.js
```diff
@@ -0,0 +1,61 @@
+class BranchSwitch {
+  constructor($dom, store) {
+    this.$dom = $dom;
+    this.store = store;
```
nphi: it seems `store` is not used

src/main.js
```diff
@@ -129,13 +133,26 @@ $(document).ready(() => {
             if (isSidebarPinned()) toggleSidebar();
             else togglePin();
           } else if (isSidebarVisible()) {
+            // @TODO(add current working branch to local store for a certin period of time)
+            // For now this could be not a good approch because when
+            // The page is force reloaded the current branch being navigated will be lost, but for now it works.
+            if (isChangedBranch && typeof branch !== 'undefined') {
+                repo.branch = branch;
```
nphi: is it possible to save branch in memory so that it wouldn't be lost if somewhere else executes `tryLoadRepo`

src/main.js
Fahme: hmm, thanks didn't notice it :).

src/adapters/github.js
```diff
@@ -107,6 +107,16 @@ class GitHub extends PjaxAdapter {
     );
   }
 
+  // @override
+  getBranches(opts) {
+    return new Promise((resolve, reject) => {
+      this._get('/branches', opts, (err, res) => {
```
duylam: Looks like this only gets first page only https://developer.github.com/v3/repos/branches/#list-branches, either parsing HTML or iterate through Link header to get all

src/main.js
```diff
@@ -129,13 +133,26 @@ $(document).ready(() => {
...
             const replacer = ['username', 'reponame', 'branch', 'pullNumber'];
             const repoChanged = JSON.stringify(repo, replacer) !== JSON.stringify(currRepo, replacer);
             if (repoChanged || reload === true) {
               hasError = false;
               $document.trigger(EVENT.REQ_START);
               currRepo = repo;
               treeView.show(repo, token);
+              adapter
+                .getBranches({repo, token})
+                .then((branches) => {
```
duylam: Look like the rendering tree and branches run parallel because the promise. I think `adapter.getBranches` may not be good approach when the rendering tree work takes 02 job (load tree and load branches). How about changing `adapter.getRepoFromPath` to load branches and append `branches` to returned value

src/view.branch.js
```diff
@@ -0,0 +1,59 @@
+class BranchSwitch {
+  constructor($dom) {
+    this.$dom = $dom;
+    this.branches = [];
+    $dom.find('.octotree-views').on('click', '.octotree-header-branch', this._showPopUp.bind(this));
```
duylam: Just minor question: how do people know the branch label is clickable ?

src/view.branch.js
```diff
@@ -0,0 +1,59 @@
...
+
+  _handleModalClick(event) {
+    event.preventDefault();
+    const clickedBranch = $(event.target).attr('data-branch-name');
+    $(event.currentTarget).trigger(EVENT.BRANCH_SWITCHED, clickedBranch);
+    this._closePopUp();
+  }
+
+  _closePopUp() {
+    $('.modal-root').remove();
```
duylam: Can't we just cache the jquery object and hide it (display: none) since this happens multiple times ?

src/main.js
Fahme: `getRepoFromPath` always gets (master), so requesting the tree for master branch while making 2 or more request to get branches (For example in case of too many branches) seems legit or no ?

src/view.branch.js
Fahme: well was thinking of what icon could be placed there to indicate it's clickable.

src/adapters/github.js
Fahme: ahh, ok thanks, so now the current implementation request for first page only and since on every tree reload there's branches requesting we will have quite a lot of calls to endpoint, so storing the current retrieved branches will also be ideal, right ?

duylam: > so now the current implementation request for the first page only.

If you mean the `current implementation` is the fetching file list on PR then Yes. However, what we're doing here is separate feature (getting branch list) so it looks like a bug (not too much critical) when the code doesn't show all branch if the amount of over 100

src/main.js
duylam: @Fahme Agreed with your concern about using `getRepoFromPath` for loading branch doesn't look right, it's just my random thought to solve the problem about rendering on tree: the tree of file and the branch list display in different timestamp. How about moving `treeView.show(repo, token);` into the `finally` method for promise returned by `getBranches` ?

buunguyen:
Gonna close until we're clear exactly what we need to do before doing it.

Issue #770, opened 2019-08-28
Multi-directory navigation is nearly unreadable


emm035:
### Description
A recent change (I think https://github.com/ovity/octotree/pull/766) made it so the side bar is useless for multiple nested directories. I'm running into this when working with java source code, where the package names create
 an almost unreadable path when the sidebar is at a reasonable width:

![Screen Shot 2019-08-28 at 1 51 08 PM](https://user-images.githubusercontent.com/13439972/63881164-3489b800-c99d-11e9-820f-3fa8eb8a97a0.png)

Is this an option that could be exposed as a setting, or even have it just truncate the entire path, rather than each section of the path?

nphi:
@emm035  !

Thanks for reporting this. I will try to fix it ASAP.

buunguyen:
Released a fix, should be available shortly. Thanks for reporting the issue!

PR #771, opened 2019-08-29
fix: truncate file properly

nphi:
Screen Shot 2019-08-29 at 9 26 08 PM

Address https://github.com/ovity/octotree/issues/770


buunguyen:
I think you should change in this https://github.com/ovity/octotree/blob/v3/src/view.tree.js#L113

PR #790, opened 2019-09-08
Add disclaimer message for personal token usage.

Fahme:
### Problem
Disclaimer message for personal token usage in Octotree is not shown in settings view  
### Solution
Adding a disclaimer message to make it clear token is stored in local storage.
### Screenshots
Screen Shot 2019-09-08 at 5 27 30 PM

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -353,6 +353,11 @@
           height: 60px;
         }
 
+        .octotree-disclaimer-token {
+          font-size: 0.8em;
+          margin-top: 6px;
+        }
+
         .octotree-disclaimer {
           color: @gray;
           display: block;
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -350,17 +350,14 @@
           resize: vertical;
           height: 60px;
         }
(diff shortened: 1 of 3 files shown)


src/adapters/github.less
```diff
@@ -353,6 +353,11 @@
           height: 60px;
         }
 
+        .octotree-disclaimer-token {
```
buunguyen: Why add a new selector? We already have .octotree-disclaimer. Can we use it (maybe customer so it is usable in both places?

src/template.html
```diff
@@ -56,6 +56,7 @@
               </a>
             </div>
             <input type="text" class="form-control input-block" data-store="TOKEN" />
+            <div class="octotree-disclaimer-token text-gray text-mono">The access token is stored in your browser local storage and Octotree uses it only to communicate with GitHub API.</div>
```
buunguyen: Please don't use GitHub classes (text-gray text-mono) for simple case like this. Just define explicitly the CSS classes.

Change text to "Token is stored in local storage"

PR #793, opened 2019-09-11
fix tooltip settings

nphi:
Screen Shot 2019-09-11 at 7 12 06 AM


buunguyen:
Should it show top or bottom? Left/right looks kinda weird. And not sure what's the reason for this change.

PR #798, opened 2019-09-13
refactor: remove installing warning

nphi:


nphi:
a @buunguyen , thanks for reminding that.

PR #806, opened 2019-10-02
Fix merge v4

nphi:


src/template.html
```diff
@@ -56,7 +56,6 @@
               </a>
             </div>
             <input type="text" class="form-control input-block" data-store="TOKEN" />
-            <div class="octotree-disclaimer text-gray">Token is stored in local storage.</div>
```
nphi: a @buunguyen , do we need this ? I'm not sure If it's correct if I remove it

buunguyen: We need it.

PR #830, opened 2019-10-24
Fix passive npm audit vulnerabilities

sharynneazhar:
### Problem
Minor updates. Found vulnerabilities upon running `npm install`.

### Solution
Ran npm audit fix to resolve vulnerabilities - 131 vulnerabilities resolved.

Note: This is a passive change. There are unresolved vulnerabilities that will potentially cause breaking changes. One of them is upgrading Gulp v3 to v4.

### Screenshots
Before: 
Screen Shot 2019-10-24 at 3 12 23 PM

After: 
Screen Shot 2019-10-24 at 3 13 10 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "octotree",
-  "version": "3.0.0",
+  "version": "3.0.10",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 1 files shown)


buunguyen:
Thanks for this. We don't have time to regression test now so not going to change dev packages.

PR #835, opened 2019-10-29
Change the global storage

endyquang:


src/adapters/github.js
```diff
@@ -315,21 +315,23 @@ class GitHub extends PjaxAdapter {
 
     $.ajax(cfg)
       .done((data, textStatus, jqXHR) => {
```
danhnguyen30041975: Why to use anonymous execution ? This doesn't work ?

```suggestion
      .done(async (data, textStatus, jqXHR) => {
```

src/main.js
```diff
@@ -2,6 +2,8 @@ $(document).ready(() => {
   octotree.load(loadExtension);
 
   async function loadExtension(activationOpts = {}) {
+    window.extStore = await ExtStore.create(STORE, DEFAULTS);
```
danhnguyen30041975: Bookmark

src/view.tree.js
```diff
@@ -30,31 +30,33 @@ class TreeView {
     const $jstree = this.$jstree;
 
     $jstree.settings.core.data = (node, cb) => {
```
danhnguyen30041975: ?

```suggestion
    $jstree.settings.core.data = async (node, cb) => {
```

src/view.help.js
```diff
@@ -4,10 +4,10 @@ class HelpPopup {
     this.store = store;
   }
 
-  init() {
+  async init() {
     const $view = this.$view;
     const store = this.store;
```
buunguyen: Why this var?

src/view.help.js
```diff
@@ -4,10 +4,10 @@ class HelpPopup {
     this.store = store;
```
buunguyen: I think we should remove local `store` this because `extStore` is global. Keep code a bit saner.

src/view.options.js
```diff
@@ -67,14 +67,15 @@ class OptionsView {
...
-      ($elm, key, value, cb) => {
+      async ($elm, key, value, cb) => {
         if ($elm.is(':radio') && !$elm.is(':checked')) {
           return cb();
         }
         const newValue = $elm.is(':checkbox') ? $elm.is(':checked') : $elm.val();
         if (value === newValue) return cb();
         changes[key] = [value, newValue];
-        this.store.set(key, newValue, cb);
+        await this.store.set(key, newValue);
```
buunguyen: ditto, use extStore, get rid of this.store var

src/core.storage.js
```diff
@@ -1,44 +1,106 @@
...
+      for (const key of Object.keys(values)) {
+        await store.setIfNull(values[key], defaults[key]);
       }
+      resolve(store);
+    })
+  }
+
+  set(key, value) {
+    const payload = {[key]: value};
+    return key === STORE.TOKEN
```
buunguyen: How is the multiaccount handled? Maybe we should do it by convension, if key.contains('token') then setLocal. Then make sure the multiaccount uses something with token in name.

src/core.storage.js
```diff
@@ -1,44 +1,106 @@
...
-
-    function parse(val) {
-      try {
-        return JSON.parse(val);
-      } catch (e) {
-        return val;
+  static create(values, defaults) {
+    return new Promise(async (resolve, reject) => {
+      const store = new ExtStore();
+      for (const key of Object.keys(values)) {
```
buunguyen: May not have worthy perf impact, but generally I'd like to see this done in parallel, like

await Promise.all(keys.map((key) => store.setIfNull(...));

src/adapters/adapter.js
```diff
@@ -67,7 +67,7 @@ class Adapter {
             item.icon = type;
 
             if (type === 'blob') {
-              if (this.store.get(STORE.ICONS)) {
+              if (await this.store.get(STORE.ICONS)) {
```
buunguyen: get rid of this.store, see other comment

src/adapters/github.js
endyquang: Yep this does not work

src/view.help.js
endyquang: It's just the name you know? Or do you mean using extStore in name only?

src/view.tree.js
endyquang: same as the callback inside `done' of ajax, this function can't be an async one

src/view.help.js
buunguyen: I mean extStore now is global, we should use it everywhere. We shouldn't pass `store` object to each view like this any more.

src/core.storage.js
endyquang: the multiaccount is now using the name 'accounts', should i add a whitelist instead?

buunguyen: I'd prefer we use a convention based approach. Maybe change it to "token_accounts", then can check if startsWith("token"). That's more generic and make it easy if we add more stuff later.

src/core.storage.js
```diff
@@ -35,13 +35,11 @@ class ExtStore {
...
   static create(values, defaults) {
-    return new Promise(async (resolve, reject) => {
-      const store = new ExtStore();
-      for (const key of Object.keys(values)) {
-        await store.setIfNull(values[key], defaults[key]);
-      }
-      resolve(store);
-    })
+    const store = new ExtStore();
+    return Promise.all(Object.keys(values).map((key) => {
```
buunguyen: Less ugly version:
```
await Promise.all(...)
return store;
```

src/adapters/adapter.js
```diff
@@ -66,7 +65,7 @@ class Adapter {
             // Uses `type` as class name for tree node
             item.icon = type;
 
-            octotree.setNodeIconAndText(this, item);
+            await octotree.setNodeIconAndText(this, item);
```
buunguyen: Is this async fn?

src/adapters/github.js
buunguyen: I wonder why it doesn't work?

src/core.constants.js
```diff
@@ -4,7 +4,7 @@ const SHOW_CLASS = 'octotree-show';
 const PINNED_CLASS = 'octotree-pinned';
 
 const STORE = {
-  TOKEN: 'octotree.access_token',
+  TOKEN: 'octotree.access.local',
```
buunguyen: octotree.token.local

src/adapters/github.js
endyquang: it did not work when i tested it. I also could not find any relevant info on google. So my coclusion is it doesn't work

src/core.storage.js
endyquang: static method can't be async, so this won't do

src/adapters/adapter.js
endyquang: Yes, there is a store.get inside
https://github.com/ovity/octotree/blob/master/src/core.api.js#L81

src/view.tree.js
```diff
@@ -194,7 +195,7 @@ class TreeView {
     }
   }
 
-  syncSelection(repo) {
+  async syncSelection(repo) {
```
danhnguyen30041975: Miss update at  https://github.com/ovity/octotree/blob/cc60d85da02a5f817f1b4c222207779398f94280/src/main.js#L146

src/view.tree.js
danhnguyen30041975: Please help put comment for reason why to use anonymous execution

src/main.js
```diff
@@ -305,8 +305,8 @@ $(document).ready(() => {
     function setHotkeys(newKeys, oldKeys) {
       key.filter = () => $sidebar.is(':visible');
       if (oldKeys) key.unbind(oldKeys);
-      key(newKeys, () => {
-        if (togglePin()) treeView.focus();
+      key(newKeys, async () => {
```
danhnguyen30041975: Not sure this works, did you test ?

src/main.js
```diff
@@ -119,21 +119,21 @@ $(document).ready(() => {
...
+        await tryLoadRepo(true);
       }
     }
 
-    function tryLoadRepo(reload) {
-      const token = octotree.getAccessToken();
-      adapter.getRepoFromPath(currRepo, token, (err, repo) => {
+    async function tryLoadRepo(reload) {
+      const token = await octotree.getAccessToken();
+      await adapter.getRepoFromPath(currRepo, token, async (err, repo) => {
```
danhnguyen30041975: No need `await` for `getRepoFromPath`, the next step after calling `getRepoFromPath` is on callback.

src/core.storage.js
```diff
@@ -1,44 +1,122 @@
...
-  static create(values, defaults) {
-    const store = new Storage();
-    for (const key of Object.keys(values)) {
-      store.setIfNull(values[key], defaults[key]);
+
+class ExtStore {
+  constructor(values, defaults) {
+    this._isInit = false;
+
+    this._init = async function () {
```
danhnguyen30041975: Why not just put this as class method like

```
async _init() {

}
```

src/core.constants.js
```diff
@@ -4,7 +4,7 @@ const SHOW_CLASS = 'octotree-show';
 const PINNED_CLASS = 'octotree-pinned';
 
 const STORE = {
-  TOKEN: 'octotree.access_token',
```
danhnguyen30041975: This breaks saved token on user, is this really necessary for changing the name ?

src/main.js
endyquang: There are places calling tryLoadRepo, so either need to return the adapter.getRepoFromPath or await for it

src/core.storage.js
endyquang: To make use of the parameters values and defaults

src/core.constants.js
endyquang: I discussed with @buunguyen and came up with this naming convention. I think it worths it

src/main.js
endyquang: It works to me

src/view.tree.js
endyquang: Thanks, i totally missed this

PR #839, opened 2019-11-08
Center Github main element when sidebar pinned

Bhacaz:
### Problem
The main container of Github was not centred like it's normally be when the sidebar is pinned. I thought it feels weird.

### Solution
I adjust some css attributes to center the main Github view.

### Screenshots
**Before**
Screen Shot 2019-11-08 at 13 59 37
**After**
Screen Shot 2019-11-08 at 13 59 23


buunguyen:
Thanks for this. I have thought about this but decide the current behavior is better.

PR #849, opened 2019-12-03
Refactor load all tree option

Fahme:


src/core.api.js
```diff
@@ -87,6 +88,16 @@ class OctotreeService {
     }
   }
 
+  async _loadAllTree(context, repo) {
```
buunguyen: The name doesn't make sense. It is called "loadX" but it actually returns boolean. Maybe 'shouldLoadEntireTreeAtOnce'. Also, don't pass in 'context', just pass in 'adapter'.

PR #861, opened 2019-12-29
Dispatch `pjax:end` event on native `document`

### LINKED ISSUE #490: `pjax:end` event not dispatched on native `document`

jdreesen:
### Problem
GitHub's own pjax implementation dispatches its events directly in the DOM, while the jQuery pjax library we use here dispatches its events only within its jQuery instance. 

Because some GitHub add-ons listen to certain pjax events in the DOM it may be necessary to forward an event from jQuery to the DOM to make sure those add-ons don't break. 

Fixes #490 

### Solution
Dispatch the `pjax:start` and `pjax:end` events (for now) in the native DOM whenever they're dispatched within jQuery.

_Note that I don't forward the event details/extra parameters or whether they're cancellable at the moment, because the pjax implementations differ in this case and I could not find a case where this info is used by listeners._

### Screenshots
Before this PR when navigating through Octotree:
grafik

After the PR or when navigating though GitHub or reloading the page manually:
grafik

As can be seen on the second screenshot, Refined GitHub adds gray dots to show the whitespace and OctoLinker adds pink dots to indicate that you can navigate to the required libraries.

---

As I have no experience in browser add-on development and there's no documentation how I could build and install my local Octotree development version in my Firefox, I don't know how to verify if the solution is working. But I've done a lot of research and digging through (minified) JavaScript and am therefore pretty confident it does.

So, could you please verify this?

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -13,10 +13,13 @@ const GH_HIDDEN_RESPONSIVE_CLASS = '.d-none';
 const GH_RESPONSIVE_BREAKPOINT = 1010;
 
 class GitHub extends PjaxAdapter {
+  constructor() {
+    super(GH_PJAX_CONTAINER_SEL);
+  }
+
   // @override
   init($sidebar) {
-    const pjaxContainer = $(GH_PJAX_CONTAINER_SEL)[0];
-    super.init($sidebar, {pjaxContainer: pjaxContainer});
+    super.init($sidebar);
 
     // Fix #151 by detecting when page layout is updated.
     // In this case, split-diff page has a wider layout, so need to recompute margin.
@@ -171,11 +174,6 @@ class GitHub extends PjaxAdapter {
         cb(null, repo);
       });
     }
-  }
-
-  // @override
-  selectFile(path) {
-    super.selectFile(path, {pjaxContainerSel: GH_PJAX_CONTAINER_SEL});
   }
 
   // @override
--- a/src/adapters/pjax.js
+++ b/src/adapters/pjax.js
@@ -4,8 +4,14 @@ class PjaxAdapter extends Adapter {
     this._pjaxContainerSel = pjaxContainerSel;
 
     $(document)
-      .on('pjax:start', () => $(document).trigger(EVENT.REQ_START))
-      .on('pjax:end', () => $(document).trigger(EVENT.REQ_END))
+      .on('pjax:start', () => {
+        $(document).trigger(EVENT.REQ_START);
+        this._dispatchPjaxEventInDom('pjax:start');
+      })
+      .on('pjax:end', () => {
+        $(document).trigger(EVENT.REQ_END);
+        this._dispatchPjaxEventInDom('pjax:end');
+      })
       .on('pjax:timeout', (e) => e.preventDefault());
   }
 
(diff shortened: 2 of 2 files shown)


buunguyen:
Hi, thanks for this. Question: I'm still puzzled how Octotree breaks other extensions. I read the original issue #490 but wasn't clear why listening on events would break things. Does that stop native events from being fired or what?

buunguyen:
Also, to be clear, you didn't test it at all? Or you just didn't test it on Firefox?

buunguyen:
I think I get it now. Let me test more carefully. Thanks again for helping out!

buunguyen:
Tested. Just need to handle re-entrance otherwise there will be call stack error: https://github.com/ovity/octotree/commit/44fdb80e117aef01c30d8b50cd219953af11dc3e.

jdreesen:
Thx for your quick reaction and testing/merging/fixing this! :)

> Also, to be clear, you didn't test it at all? Or you just didn't test it on Firefox?

I didn't test it at all, because I didn't know how (I have zero experience in add-on development).

> Question: I'm still puzzled how Octotree breaks other extensions. I read the original issue #490 but wasn't clear why listening on events would break things. Does that stop native events from being fired or what?

The problem were not the listeners in Octotree, but that Octotree uses a jQuery pjax plugin, whereas GitHub uses a custom pjax implementation without jQuery that's operating directly on the DOM.

While jQuery's event listeners get called for events dispatched through jQuery *or* the DOM, everything was fine on your side.

Events dispatched through jQuery, on the other hand, are only visible within jQuery but not the DOM.

This means, that because those other plugins listen to events directly on the DOM, their listeners don't get called when Octotree dispatches its pjax events, which is why we have to re-dispatch them directly on the DOM (which... results in a loop, because jQuery's listeners get called again, so thx for your re-entrance fix!).

Here's a little example:
```js
document.addEventListener('my:test', e => console.log('DOM', e.type))
jQuery(document).on('my:test', e => console.log('jQuery', e.type))

document.dispatchEvent(new Event('my:test'))
// > DOM my:test
// > jQuery my:test

jQuery(document).trigger('my:test')
// > jQuery my:test
```

PR #866, opened 2020-01-02
Fix: Tree icons stretch in Firefox

reicheltp:
### Problem

The icons in the tree are stretched, thus the whole tree looks weird.

image

### Solution
Change `flex` property of icon to prevent stretching.

Result:
image

### Platform Info
- Octotree Version 4.1.3
- Firefox 71 (64bit) on Windows 10

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -223,7 +223,7 @@
 
         .jstree-icon {
           width: 16px;
-          flex-shrink: 0;
+          flex: 0 0 auto;
         }
 
         .jstree-icon,


jdreesen:
Strange, this doesn't occur for me with the same setup:
* Octotree version: 4.1.3
* Browser & version: Firefox 71.0 (64-Bit)
* OS & version: Windows 10 Pro 19.03

reicheltp:
Hey @jdreesen, Thanks for the tip! I investigated a little more and this strange bug was caused by an old version of Sourcegraph (1.18.0) I have also installed. Disabling this addon or updating to a newer version of Sourcegraph (19.12.x) solved the visual errors.

PR #868, opened 2020-01-03
Fix content jumps right if sidebar is pinned

endyquang:
### Problem
Due to #865  when pin, the content jumps right

### Solution
Add padding to the caculation of margin-left of the content

### Screenshots
Before #865
The content only jumps right when the sidebar is pinned and its width is big enough
a

After #865
The content jumps right when the sidebar is pinned
a

After this
The content never jumps right
a

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -76,17 +76,28 @@ class GitHub extends PjaxAdapter {
 
   // @override
   updateLayout(sidebarPinned, sidebarVisible, sidebarWidth) {
-    const SPACING = 10;
+    const SPACING = 20;
     const $containers =
       $('html').width() <= GH_RESPONSIVE_BREAKPOINT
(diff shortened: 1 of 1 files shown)


src/adapters/github.js
```diff
@@ -76,17 +76,28 @@ class GitHub extends PjaxAdapter {
 
   // @override
   updateLayout(sidebarPinned, sidebarVisible, sidebarWidth) {
-    const SPACING = 10;
+    const SPACING = 20;
```
buunguyen: Why not 10? Does it change existing behavior somehow?

buunguyen:
@endyquang I already reverted the last PR. Please issue a new PR, focusing on fixing the issue, not refactoring. We will refactor when we have the full test suite.

endyquang:
Ok let's close this. I'll make another PR

src/adapters/github.js
endyquang: Yes it changes the spacing between sidebar and the content

PR #890, opened 2020-02-23
fix New PR page layout

endyquang:
### Screenshots
Before
image

After
Peek 2020-02-23 14-48

### DIFF
--- a/src/adapters/github.js
+++ b/src/adapters/github.js
@@ -7,7 +7,7 @@ const GH_PJAX_CONTAINER_SEL =
   '#js-repo-pjax-container, div[itemtype="http://schema.org/SoftwareSourceCode"] main, [data-pjax-container]';
 
 const GH_CONTAINERS = '.container, .container-lg, .container-responsive';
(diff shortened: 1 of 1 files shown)


buunguyen:
@endyquang does this also address these issues? https://github.com/ovity/octotree/issues/902

endyquang:
@buunguyen No this doesn't. This will only fix the PR page

buunguyen:
Can you check the linked issue too, see if can change to fix it so that we only regression once?

endyquang:
Certainly this PR should fix every possible pages. I had checked the 902 and could not find any quick solution, i'll try some more

buunguyen:
In the GIF, why did the margin changed when switching from .gitignore to gulpfile.js? Also when switching from the diff page to code page, margin also changes.

endyquang:
@buunguyen it's because of the scrollbar of the browser. The scrollbar pop out for gulpfile and dissapear when open .gitignore.
I suppose a Mac does't have this, but this is a normal behavior with linux and windows

Issue #895, opened 2020-03-03
File links in pull requests are broken


RussKie:
### Description

The file navigation is broken after the recent GitHub update.
Seems `id`s for files have changed. E.g. it used to be:
```html
<div id="diff-6" class="file ...." data-codeowners="...." data-file-type=".cs" data-file-deleted="false">
   ....
</div>
```
and now it is
```html
<div id="diff-56f8482aeb786dd105fbf93dedf2615f" class="file ...." data-codeowners="..." data-file-type=".targets" data-file-deleted="false">
   ....
</div>
```

[Better Pull Requests](https://github.com/berzniz/github_pr_tree/pull/96) fixed it pretty fast 😉 

### Environment (if bug)

* Octotree version: 4.1.7
* Browser & version: MS Edge Version 80.0.361.62 (Official build) (64-bit)
* OS & version: W10

buunguyen:
Thanks, we're trying to fix this. The Better Pull Requests folks are lucky, they only show the tree in the "Files" tab, where the new DOM is available so the fix is very simple. Octotree needs to show in other pull request pages (e.g. Conversations, Commits), where the new hashes aren't known, we're still trying to figure out where those come from (thanks GitHub!).

buunguyen:
This was quite challenging to fix, as pointed out in the last comment, but also because we had to handle backward compatibility with old versions of GitHub Enterprise. We decided to move the pull request review feature to the Pro version so that contributors could be **partially** paid for their time. This is the only way we can keep Octotree alive. The fix is available in 4.2.1. 

As a token of appreciation for those of you who reported the issue and discussed about it in this thread and [the related thread](https://github.com/ovity/octotree/issues/905), we offer a lifetime subscription to Octotree Pro. Please [contact us](https://www.octotree.io/contact) with the email address of your Octotree account.

RussKie:
It is an unfortunate loss of functionality, but I totally respect your decision. 
Thank you for the generous offer 👍

PR #903, opened 2020-03-15
Remove PR changes view

endyquang:
### Problem
https://twitter.com/furixturi/status/1233228545125519360
#895 
Github has changed. The files navigation is broken, it no longer links to the correct files and works as we expected.

### Solution
Fixing this would be too overhead and would add so much complexity to both code and usability. So the solution is to **remove this feature**.

### DIFF
--- a/src/view.tree.js
+++ b/src/view.tree.js
@@ -99,25 +99,12 @@ class TreeView {
   }
 
   _onItemClick(event) {
-    let $target = $(event.target);
-    let download = false;
-
     // Handle middle click
     if (event.which === 2) return;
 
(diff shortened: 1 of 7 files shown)


src/view.tree.js
```diff
@@ -99,25 +99,12 @@ class TreeView {
   }
 
   _onItemClick(event) {
-    let $target = $(event.target);
-    let download = false;
```
endyquang: Beside from adding a new method to detect the click target.
I removed this download part, which view the file's raw. So as not to conflict with #892

Issue #904, opened 2020-03-16
Settings view can't be scrolled


gabrielecanepa:
### Description

The settings view can't be scrolled due to an overflow issue. The problem can be fixed with an `overflow-y: scroll`, but this causes the tooltip to be cut off beyond the boundary (see https://github.com/ovity/octotree/pull/906).

### Environment

* Octotree version 4.1.7
* Chrome version 80.0.3987.132
* macOS Catalina 10.15.3

### Screenshots

![screenshot1](https://user-images.githubusercontent.com/24722181/76749382-b3e5aa00-677c-11ea-9aa7-b85d0fed22ec.gif)

PR #906, opened 2020-03-16
Fix scroll in settings view

gabrielecanepa:
### Description

Fixes https://github.com/ovity/octotree/issues/904.

### DIFF
--- a/src/adapters/github.less
+++ b/src/adapters/github.less
@@ -313,6 +313,8 @@
     }
 
     .octotree-settings-view {
+      overflow-y: scroll;
+
       & > form {
         padding: 8px;
 


buunguyen:
Thanks for this. Sadly, this causes the tooltip to be cut off beyond the boundary. Wonder if that can be fixed.

gabrielecanepa:
Right, I forgot to check that. I'll try to find another solution 👍

gabrielecanepa:
I'm closing this pr since I unfortunately don't have any time to work on this bug now. The issue will remain.
Thanks for your amazing work, I'll propose new fixes/feature and try to help whenever I can as I keep using the app!

Issue #912, opened 2020-03-27
When API rate limit is reached, message indicating that "private repositories requires a GitHub access token" is unclear


grantwinney:
### Description
I hit the API limit, since I hadn't generated an access token.

The console output indicates as such, although admittedly I have a number of GitHub related extensions enabled at the moment, and any of them could be throwing this message:

> Failed to load resource: the server responded with a status of 403 (rate limit exceeded)

The Octotree panel just shows "Error: Forbidden" at the top, with a body of:

> Accessing private repositories requires a GitHub access token. Please go to Settings and enter a token.

I was confused about this for a few minutes because I wasn't trying to access private repos. It might help, at the very least, to change [the message to something like](https://github.com/ovity/octotree/blob/856a4bd66eacf18403fba9f69bfef287c5f79adb/src/adapters/adapter.js#L170):

> A GitHub access token is required to access private repositories, or when the API rate limit is exceeded. Please go to Settings and enter a token.

### Environment (if bug)

* Octotree version: v20.3.25.911
* Browser & version: Brave, Version 1.5.115 Chromium: 80.0.3987.149 (Official Build) (64-bit)
* OS & version: Windows 10 Pro
* Screenshot, if any (drag an image here)
* Console error log below this line

PR #913, opened 2020-03-29
Change text in error forbidden tab

Fahme:
### Problem
fixes https://github.com/ovity/octotree/issues/912

### DIFF
--- a/src/adapters/adapter.js
+++ b/src/adapters/adapter.js
@@ -157,7 +157,7 @@ class Adapter {
           'Please go to <a class="settings-btn">Settings</a> and enter a token.';
         break;
       case 403:
-        if (jqXHR.getResponseHeader('X-RateLimit-Remaining') === '0') {
+        if (jqXHR.statusText === 'rate limit exceeded') {
           // It's kinda specific for GitHub
           error = 'API limit exceeded';
           message =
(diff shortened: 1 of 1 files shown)


src/adapters/adapter.js
```diff
@@ -167,7 +167,7 @@ class Adapter {
         } else {
           error = 'Forbidden';
           message =
-            'Accessing private repositories requires a GitHub access token. ' +
+            'A GitHub access token is required to access private repositories, or when the API rate limit is exceeded. ' +
```
buunguyen: You know when this case is activated? I mean I thought private repo should hit the 'case 404' and api limit should hit the first if block of 'case 403'. If those are not handled correctly, should we update them?

Fahme: This check is activated in these cases : 
- GH rate limit exceeded and server responds with 403.
- GH blocks a certain IP and server responds with 403.

So yes there's a bug and it's due to the check at line 160 is wrong ```jqXHR.getResponseHeader('X-Ratelimit-Remaining')``` 
this line returns null when rate limits is exceeded. The function is returning null maybe (due to some issue with CROS).

If can't find a way to read the rate limit header for now first thought i have in mind is to currently go with the suggestion described in the issue and keep a generic error message for 403 case like this : 

```A GitHub access token is required to access private repositories, or when the API rate limit is exceeded.```

buunguyen: Can we detect the response to fix the root cause instead? So that it shows the full error message in line 164 (which explains what rate limit is).

PR #927, opened 2020-04-22
fix: don't render on sponsors pages

zacanger:
### Problem
Octotree renders on Sponsors pages, which are not repos.

### Solution
Add sponsors to list of reserved names.

### Screenshots
Before:
2020-04-22-123023_415x263_scrot

I don't have an after screenshot because this is using Gulp 3, and I'm on Node 13: https://github.com/gulpjs/gulp/issues/2324 so I can't build and test.


buunguyen:
Thank you!

PR #932, opened 2020-04-23
fix: fix header height error at large layout

ricosmall:
### Problem

When the screen width bigger than 1012px, height of header is wrong.

### Solution

Add new variable and media query to solve the problem.

### Screenshots

**Before**

image

**After**

image


buunguyen:
Thank you!

PR #935, opened 2020-04-23
Add check for unset storages values to #934

Fahme:
This is a follow up fix for #934


src/core.storage.js
```diff
@@ -14,7 +14,7 @@ class ExtStore {
     this._init = Promise.all(
       Object.keys(values).map(async (key) => {
         const existingVal = await this._innerGet(values[key]);
-        if (existingVal === null) {
+        if (existingVal === undefined || existingVal === null) {
```
buunguyen, line 17: Just put `existingVal == null`, that's more idiomatic.

PR #966, opened 2020-06-08
Better icon match for Codecov

CismonX:
### Problem

Icon of Codecov configuration file `.codecov.yml` appears to be the icon for a generic YAML file instead of Codecov's, for octotree only matches `codecov.y(a)ml` but not `.codecov.y(a)ml`.

Codecov permits both names, as said in [the official documentation](https://docs.codecov.io/docs/codecov-yaml#can-i-name-the-file-codecovyml):

> **Can I name the file .codecov.yml?**
> Yes. However, the file must still be located in the root, `dev/`, or `.github/` directories

### Solution
Update the icon database so that octotree can associate both names with Codecov.

### Screenshots

Before:
before

After:
after


buunguyen:
Thanks for this. This is 3rd-party lib, not our code: https://github.com/websemantics/file-icons-js. I'd prefer that lib is updated. It doesn't seem to be maintained, maybe an opportunity to request to maintain it?

